PR #15, opened 2016-05-07
Fixing nested Transaction bug with Sqlite.

Skrypt:
I have tested with the YesSql.Samples.Hi project.
I've found that now we need to specify in the connection string "Cache=Shared" if we are using a db file.
I also noticed that Sqlite by default should have : 

```C#
cfg.IsolationLevel = IsolationLevel.Serializable
```

P.S. Needs more testing before merging.

### DIFF
--- a/test/YesSql.Tests/project.json
+++ b/test/YesSql.Tests/project.json
@@ -19,7 +19,7 @@
       "dependencies": {
         "NETStandard.Library": "1.5.0-*",
         "dotnet-test-xunit": "1.0.0-dev-*",
-        "System.Reflection.Extensions": "4.0.1-rc2-23931",
+        "System.Reflection.Extensions": "4.0.1-rc2-24027",
         "System.Runtime.Serialization.Primitives": "4.1.1-rc2-23931"
       }
     },
--- a/test/YesSql.Tests/project.json
+++ b/test/YesSql.Tests/project.json
@@ -18,19 +18,9 @@
       ],
       "dependencies": {
         "NETStandard.Library": "1.5.0-*",
-        "dotnet-test-xunit": "1.0.0-dev-*",
+        "dotnet-test-xunit": "1.0.0-rc2-*",
         "System.Reflection.Extensions": "4.0.1-rc2-24027",
-        "System.Runtime.Serialization.Primitives": "4.1.1-rc2-23931"
-      }
-    },
-    "net451": {
-      "frameworkAssemblies": {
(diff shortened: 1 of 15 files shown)


sebastienros:
Have you tried to run all unit tests with the `Demand` call removed ?

Skrypt:
This is what I'm getting when running dotnet test directly from the command shell.
image

test/YesSql.Tests/project.json
```diff
@@ -19,7 +19,7 @@
       "dependencies": {
         "NETStandard.Library": "1.5.0-*",
         "dotnet-test-xunit": "1.0.0-dev-*",
```
alexbocharov: @sebastienros @Skrypt It isn't correct 
`"dotnet-test-xunit": "1.0.0-rc2-*"` or `"dotnet-test-xunit": "1.0.0-rc2-173384-19"` 

More information: [aspnetcidev feed](http://myget.org/feed/aspnetcidev/package/nuget/dotnet-test-xunit) or [aspnetcirelease feed](http://myget.org/feed/aspnetcirelease/package/nuget/dotnet-test-xunit)

Skrypt:
I'm getting the same result.

test/YesSql.Tests/project.json
Skrypt: I'm getting the same result after changing it to either one of those.

Skrypt: I think it is related to https://github.com/aspnet/KoreBuild/pull/64

alexbocharov: You try to remove `.nuget/packages`. After that `dotnet restore`.

Skrypt: I removed these lines in the YesSql.Tests project.json file.

```
    "net451": {
      "frameworkAssemblies": {
        "System.Runtime": {
          "type": "build"
        },
        "System.Reflection": {
          "type": "build"
        },
        "System.Threading.Tasks": {
          "type": "build"
        },
        "System.Runtime.Serialization": "4.0.0.0"
      },
      "dependencies": {
      }
    }
```
And ran tests succesfully : 

image
image


So this is related to : https://github.com/aspnet/KoreBuild/pull/64

sebastienros:
Why did you comment on the commits and not the PR ;) But still, if all the tests pass after that then that's awesome, I'd really like to understand what I intended with a call to `Demand` there.

Issue #39, opened 2017-03-28
Prettify the Data Access APIs


hishamco:
Based on the discussion [here](https://github.com/OrchardCMS/Orchard2/issues/638) with @sebastienros in the Orchard CMS repo , there must be changes in both repos to support the suggested APIs, so the developer can simply use YesSQL data provider(s) as the following:

```csharp
public void ConfigureServices(IServiceCollection services)
{
    services.AddDbProvider(options =>
        options.UseSqlServer("Server=.;Database=YesSqlDb;Integrated Security=True"));
    ...
}
```

So, YesSql needs to be configured with a `DbFactory` instance containing a connection string, which you are showing correctly in the example above.

My initial thoughts here is:

- Create the following projects named `YesSql.Provider.InMemory`, `YesSql.Provider.Sqlite`, `YesSql.Provider.SqlServer`, `YesSql.Provider.MySql`, `YesSql.Provider.Postgres`, to support all the data providers which supported by YesSQL, so the user can reference only the one that he interest with (Pluggability)

- Add required unit tests for the suggested APIs

- Provide a web sample as a show case `YesSql.Samples.Web`

@sebastienros please let me know if you have any further suggestion

sebastienros:
`AddDbFactory` in this case? 
Also if a DbFactory accepts a delegated one it should be reflected. For instance https://github.com/sebastienros/yessql/blob/master/src/YesSql.Storage.Cache/CacheDocumentStorageFactory.cs#L15

hishamco:
`AddDbFactory` instead of `AddDbProvider`?

> Also if a DbFactory accepts a delegated one it should be reflected. For instance https://github.com/sebastienros/yessql/blob/master/src/YesSql.Storage.Cache/CacheDocumentStorageFactory.cs#L15

+1, also we can do this by introducing a caching option in the APIs

```csharp
services.AddDbProvider(options =>
        options.UseSqlServer("Server=.;Database=YesSqlDb;Integrated Security=True",
        cached: true));
```
so in this case `newCacheDocumentStorageFactory(new SqlDocumentStorageFactory())` will be created instead of `new SqlDocumentStorageFactory()`

PR #40, opened 2017-03-30
Prettify Data Access APIs

hishamco:
Fixes issue #39


hishamco:
@sebastienros why I got this message *Build worker image not found: Visual Studio 2017 RC*? It makes AppVeyor failed!!

sebastienros:
Can you rebase on master, I fixed the CI

src/YesSql.Core/Provider/IDbProviderOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using YesSql.Core.Services;
+
+namespace YesSql.Core.Provider
+{
+    public interface IDbProviderOptions
+    {
+        string ProviderName { get; set; }
```
sebastienros: I don't really see the value of this property. Hence we could just remove the `IDbPRoviderOptions` class and use `Configuration` directly in the `AddDbProvider` method.

src/YesSql.Provider.InMemory/InMemoryDbProviderOptionsExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+            this IDbProviderOptions options)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            var configuration = new Configuration
+            {
+                ConnectionFactory = new DbConnectionFactory<SqliteConnection>(ConnectionString, true),
```
sebastienros: Interesting idea, use the in-memory storage with an in-memory sqllite index. Didn't think about it.

src/YesSql.Provider.MySql/MySqlDbProviderOptionsExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+            if (String.IsNullOrWhiteSpace(connectionString))
+            {
+                throw new ArgumentException(nameof(connectionString));
+            }
+
+            var configuration = new Configuration
+            {
+                ConnectionFactory = new DbConnectionFactory<MySqlConnection>(connectionString, true),
+                DocumentStorageFactory = new SqlDocumentStorageFactory(),
+                IsolationLevel = IsolationLevel.ReadUncommitted
```
sebastienros: Maybe we should expose this in an `UseMySql` overload, and for the other ones that can use such a value.

sebastienros:
I like it, it seems to make things much simpler to use. 👍

src/YesSql.Provider.MySql/MySqlDbProviderOptionsExtensions.cs
hishamco: I already made a generic `Use` that expose the `IsolationLevel` and `DocumentStorageFactory`, but it good to expose `IsolationLevel` if there's a value for this, coz I saw the default implementation in your code using 'ReadUncommitted` except the Sqlite

hishamco:
@sebastienros I already rebase, but still, seems there's an issue with `MSSQL$SQL2014` service!!

src/YesSql.Core/Provider/IDbProviderOptions.cs
hishamco: May be, I'm still thinking for different scenarios one of them is `Orchard CMS` where such property has value to determine the default database provider

sebastienros: No, Orchard CMS works at a different level, once it knows what DbProvider and connection string to use, then it can call UseXXXServer(connectionString). It needs a list of available providers before this call. This is why Orchard "CMS" has its own providers extensibility.

hishamco:
Hope you fix the CI issue again

src/YesSql.Core/Provider/IDbProviderOptions.cs
hishamco: IMHO once it knows the `ProviderName` it can use the appropriate data provider

sebastienros:
I did some work on the CI, it fixed SQL Server (there is an issue with appveyor I think) but the PostgreSql tests don't work anymore. I will look into it, it's independent from this PR, I will check on master.

hishamco:
I see, because I already recognized that some tests are failed, so shall I rebase or this will remove your changes?!! I think we can fix this in another PR , and I will rebase after that

src/YesSql.Core/Provider/IDbProviderOptions.cs
sebastienros: You don't seem to understand, so let me give you a more detailed explanation.

When we start Orchard CMS, we display a list of a supported database names: 
1- Sqlite, Sql Server, Postgres, MySql. 
2- The user selects one from this list.
3- We then show or hide the Connection String text input, and the Table Prefix one.
4- Once the user has entered these information we save it in configuration. 
5- The application is then reloaded and we can read the configuration to call UserSqlServer(cstr) or UseSqlite(cstr) ...based on what the user chose.

How do you think having a provider name in this class will help us list all the available providers on the step 1? I can tell you there is none. At that point we can't call this option, because we don't know what the user wants. Also where do you specify that a provider accepts connection strings or not? Again, not in this class. What exists already in Orchard CMS is both necessary and sufficient for listing the providers.

Now with your other changes, we can call UseXXX based on the configuration Orchard CMS will store.

sebastienros:
I'll merge when we get rid of the custom options. Bonus point for overloads on IsolationLevel.

sebastienros:
If you rebase on master the tests should pass now.

src/YesSql.Core/Provider/IDbProviderOptions.cs
hishamco: Thanks for detailed explanation, if that's the case I think Orchard CMS is already have a way to list those providers from DI or some where else, which is out of the scope here

But still I see there's a benefit to have a `IDbProviderOptions`, perhaps the 'ProviderName` is not that useful here, but at least the use can know the current provider without digging into the `DbConnectionFactory`, also `IDbProviderOptions` may useful to keep track of table prefix

> Also where do you specify that a provider accepts connection strings or not?

this is done in the extension methods `UseSqlServer("connectionString")`

hishamco:
I will submit the `IsolationLevel` overload after few minutes

src/YesSql.Core/Provider/IDbProviderOptions.cs
sebastienros: > if that's the case I think Orchard CMS is already have a way to list those providers from DI or some where else, which is out of the scope here

Yes, so you can remove the `ProviderName` option

> > Also where do you specify that a provider accepts connection strings or not?

> this is done in the extension methods UseSqlServer("connectionString")

You quote me, but you missunderstand this quote. Yes the extension method accepts a connection string. But in Orchard CMS if a provider doesn't need a connection string to be provided by the user we don't display the text box for it in the UI. I am talking "UI metadata" not actual method call.

hishamco: I got you now 😄, in that case we can add another extension that expose `UseXXX` without providing connection string and supply it from the settings inside the newly created method

hishamco: Look @sebastienros while we will remove the `ProviderName` property, so the `IDbProviderOptions` will be useless, let us remove it for now, and if we there's a need in the future to have provider name, table prefix .. ect, we will bring it back.

Are you agree with that?

sebastienros: image

First comment of the discussion. Yes, because this property is not necessary, this class isn't either. Even more we already have a class with all the necessary properties, including `TablePrefix`. It's called `Configuration`. So we just need to have all these extension methods and new projects use it instead of `IDbProviderOptions`.

hishamco: Ok, last thing, if I understand you correctly `UseXXX` without connection string is Orchard CMS specific thing, so should be added there

src/YesSql.Provider.MySql/MySqlDbProviderOptionsExtensions.cs
```diff
@@ -27,7 +35,7 @@ public static class MySqlDbProviderOptionsExtensions
             {
                 ConnectionFactory = new DbConnectionFactory<MySqlConnection>(connectionString, true),
```
sebastienros: This should use the default value for `reuseConnection`. Meaning you can remove the `true`. Do this on all extension methods.

hishamco: Sure, I will update them now ..

hishamco: There's only one constructor https://github.com/sebastienros/yessql/blob/b6f5e719a1c2147d599c705bcb23057561e32f94/src/YesSql.Core/Services/Configuration.cs#L42 that requiring `reuseConnection`, so how we can remove it?!!

src/YesSql.Core/Provider/IDbProviderOptions.cs
sebastienros: No. The user doesn't specific a connection string in the UI based on the UI metadata, but Orchard CMS will provide one. Don't change anything.

src/YesSql.Provider.MySql/MySqlDbProviderOptionsExtensions.cs
sebastienros: The default value is false. Remove `, true` and it will use it. Example:
https://github.com/sebastienros/yessql/blob/master/test/YesSql.Tests/SqlServerTests.cs#L17

hishamco: Yep optional param .. sorry it's my bad 😄

sebastienros:
One test is failing, once it's fixed I can merge if you are satisfied with it.

hishamco:
After downloading the .NET Core latest bits, I faced some issues with VS 2017

The test should passed now

PR #58, opened 2017-05-25
Fix YesSqlProvider namespace

hishamco:


hishamco:
@sebastienros could you please accept this fix?

Issue #56, opened 2017-05-25
InMemoryDbProvider is missing


hishamco:
After that latest updates, seems the `InMemoryDbProvider` has been deleted, it was a good choice specially for testing purpose is there a reason for the deletion?

sebastienros:
You can already do everything in memory using the Sqlite provider and using an in-memory connection string.
I will reintroduce technical providers (cache, hub, audit) later, by implementing custom IStore and ISession.

The idea will be that a CacheStore and CacheSession will proxy/wrap an actual store and session. Anything that can be fulfilled by these will return, otherwise the command will be passed to wrapped instance. 

#### Cache
Will prevent Get(int) from hitting the database if the instance is already known. Can also cache some queries and invalidate them on any write.

#### Hub
A composite for other sessions. For instance you can register two session on a hub, one for Sqlserver and on for Sqlite. Every write would go to both, reads would go to the main one. This allows replication of data. 

#### Audit
Like an event store. Records all actions in a storage format. You can then look at the history of changes on object, and also implement synchronization between systems, for instance by developing disconnected application that would synchronize the changes with the main system once reconnected.

hishamco:
thanks for detailed information, but it would be nice if we have `InMemoryDbProvider` out-of-the-box instead of remembering the connection required for the SqLite to use the in memory data store

sebastienros:
I see what you mean. Feel free to make a pr and add it as another extension method in the sqlite provider.

sfmskywalker:
It appears the `InMemory` provider isn't available in the `dev` branch.

So I tried using the `Sqlite` provider's in-memory connection string (`"Filename=:memory:"`), which in itself works, except for the following issue: every time the connection goes out of scope, all of the created tables are lost from memory as well.

To work around this, I can create a custom DbConnectionFactory that always returns the same connection (effectively being a singleton). I tried this, and that seems to work nicely.

I can create a PR which adds support for in-memory Sqlite, unless we think we should re-introduce the `InMemoryDbProvider`?

hishamco:
> It appears the InMemory provider isn't available in the dev branch.

Seems Seb remove it long time back ;)

sfmskywalker:
Yeah, he moved that cheese way back ;)

In any case, here is a working YesSql provider using Sqlite's in-memory capability: https://github.com/elsa-workflows/elsa-core/tree/feature/elsa-2.0/src/core/YesSql.Provider.Sqlite.InMemory

[...]

PR #59, opened 2017-05-25
Use InMemory data storage on top of SqLite

hishamco:
Fixes issue #56 

/cc @sebastienros

### DIFF
--- a/test/YesSql.Tests/InMemoryTests.cs
+++ b/test/YesSql.Tests/InMemoryTests.cs
@@ -1,0 +1,50 @@
+using System.Threading.Tasks;
+using Xunit;
+using YesSql.Provider.Sqlite;
+using YesSql.Sql;
+
+namespace YesSql.Tests
+{
+    public class InMemoryTests : CoreTests
+    {
+        public InMemoryTests()
+        {
+            _store = new Store(new Configuration().UseInMemory());
+
+
+            CleanDatabase();
+            CreateTables();
+        }
+
+        protected override void OnCleanDatabase(ISession session)
+        {
+            base.OnCleanDatabase(session);
+
+            var builder = new SchemaBuilder(session);
+
+            try
+            {
+                builder.DropTable("Content");
+            }
+            catch { }
+
+            try
+            {
+                builder.DropTable("Collection1_Content");
+            }
+            catch { }
+        }
+
+        [Fact(Skip = "ReadCommitted is not supported by Sqlite")]
+        public override Task ShouldReadCommittedRecords()
+        {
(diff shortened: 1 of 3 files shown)


sebastienros:
Thank you. Can you also add a test class like the Sqlite one?
Also, the change doesn't compile (see CI)

sebastienros:
When you add the test class, also update the appveyor.yml file to opt in for this test. Look at the end of the file, this will be obvious.

hishamco:
There was a tiny issue before, I fix it and added the test too, nothing but the CI shows an error after adding the test

Any Idea?

sebastienros:
I will test this locally, we might need to set the ReuseConnection argument to true, or the data is lost after each session.

hishamco:
Strange!! It was working with the same `ConnestionString` before, have a look to the old commit https://github.com/sebastienros/yessql/commit/1d643b1f89b60592e1e6e2852bd31dc0b46e688d

sebastienros:
The _new_ unit test class is failing.

hishamco:
I'm getting **no such table: Identifiers** any idea?!!

sebastienros:
I just pushed the fixes that you need. I tried to create it myself and all tests pass.

You just have to rebase on master, then add your extension method and don't forget to pass `true` for `sharedConnection`. Then it will work.

hishamco:
Well done!! thanks @sebastienros

test/YesSql.Tests/InMemoryTests.cs
```diff
@@ -0,0 +1,50 @@
...
+            catch { }
+
+            try
+            {
+                builder.DropTable("Collection1_Content");
+            }
+            catch { }
+        }
+
+        [Fact(Skip = "ReadCommitted is not supported by Sqlite")]
```
sebastienros: Have you tried? Maybe it works in memory.

test/YesSql.Tests/InMemoryTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        public override Task ShouldReadCommittedRecords()
+        {
+            return base.ShouldReadCommittedRecords();
+        }
+
+        [Fact(Skip = "Sqlite doesn't support concurrent writers")]
```
sebastienros: Same comment here

test/YesSql.Tests/InMemoryTests.cs
hishamco: The test is failed 😞

hishamco:
I update the PR to make `sharedConnection` as optional parameter

Issue #63, opened 2017-06-09
Fix Paging Support in SQL Server 2008


hishamco:
There's a bug in paging in previous versions of SQL Server, `OFFSET` is supported on SQL Server 2012

FYI this issue is already logged [here](https://github.com/OrchardCMS/Orchard2/issues/725)

PR #64, opened 2017-06-09
Support paging in SQL Server 2008 R2

hishamco:
Fixes issue #63


sebastienros:
Not sure I really like the idea of using a syntax from 2008 as it as an impact on performance: https://www.mssqltips.com/sqlservertip/2696/comparing-performance-for-different-sql-server-paging-methods/

I would prefer an option where we have a distinct dialect, and all the changes are done in the dialect.
An option could be to add a parameter in the connection string and detect it when looking up the dialect. Another way would be to set the dialect explicitly in the configuration.

hishamco:
I didn't know that will impact the performance
Your idea is great, but let me try something else, instead of introducing a new dialect

hishamco:
I was thinking to get the Server name using `@@VERSION` or `SqlConnection.ServerVersion` in `UseSqlServer` after that we can easily use the suitable code for each version, or much simpler we can add an optional parameter name `supportPriorSQLServer2012` which is `false` by default

sebastienros:
In the end you will still need to switch the code to some other implementation, and a dialect, even inheriting from the 2016 one would be nice. This is what NHibernate does for instance.

hishamco:
So if the optional param is fine, I will add it, also I will create another dialect

sebastienros:
Parameter is fine. This should not be limited to SqlServer but general to any configuration IMO.

For the SqlServerBuilder I disagree, if we need a special case then it means it's in dialect, and then we have to add another method to special the processing. Look at "InOperator" for instance which is special for Postgre.

hishamco:
> This should not be limited to SqlServer but general to any configuration IMO

Do you mean passing dialect?!!, because I suggest to add `supportPriorSQLServer2012` which is suited for SQL Server case

sebastienros:
Yes I mean passing the dialect. You can add an extension method that does it with the correct dialect if you want too.

But please check wherever we currently load the dialect, and how, to be sure we will always take the one specified. I know that there is a static array for now, we should only use it to initialize the value if not set by configuration.

hishamco:
Ok, let me have a try ..

hishamco:
I pushed the PR #65 to add the dialect as optional param to `UsingXXX`, this PR one will be on top of the previous one

DaveEmmerson:
@hishamco @sebastienros I've just hit this myself when trying to deploy to an old machine. I'm aware that this has been idle for 2 years now, so might be a bit out of date, but it's exactly what I was about to raise an issue for..., I was even going to suggest a copy of the SqlServer dialect with just the paging part changed.

Is this something that's still viable?

sebastienros:
There is a way to pass custom dialects to the configuration now, @hishamco do you want to update the PR to simply create a new dialect class. Maybe another test class that inherits from SqlServerTests can override the sqldialect instance and runs all the tests with this new one? As long as the current instance used for unit tests will handle that.

hishamco:
I will try to create a fresh PR, or I will update this one

deanmarcussen:
@hishamco could you merge dev on this with the key length tests?

My concern is that it will now fail because of key length limitations on the earlier version of sql server.

The other command that will fail (but does not have a test for it) is the `DropIndex` command

And those are only the two I know about.

I'm not against supporting the older version, but I'd be wary that there are not more issues or limitations that are hard to work around.

hishamco:
Ok Dean, I will merge dev on this, but as you said it should run against SQL Server 2008 to make sure it works 100%

src/YesSql.Provider.SqlServer/SqlServer2008R2Dialect.cs
```diff
@@ -0,0 +1,22 @@
...
+            if (limit != null)
+            {
+                var selector = sqlBuilder.GetSelector();
+                selector = " top " + limit + " " + selector;
+                sqlBuilder.Selector(selector);
+            }
+
+            if (offset != null)
+            {
+                sqlBuilder.WhereAnd("RowNum >" + offset);
```
jersiovic, line 18: What is RownNum? I've tested a query using it like the one yessql would produce but it returns an error:
`SELECT DISTINCT TOP 10  [Document].*, [UserIndex].[NormalizedUserName] FROM [Document] INNER JOIN [UserIndex] ON [UserIndex].[DocumentId] = [Document].[Id] WHERE RowNum > 0 ORDER BY [UserIndex].[NormalizedUserName]`

This is the error I receive `'Msg 207, Level 16, State 1, Line 27
Invalid column name 'RowNum'`
I just ask cause I'm woriking on a new dialect supporting SqlServer2008 but the only way I found to do pagination produces a query like that:
SELECT * FROM (SELECT DISTINCT [Document].*, ROW_NUMBER() OVER( ORDER BY [UserIndex].[NormalizedUserName]) AS Seq,[UserIndex].[NormalizedUserName] FROM [Document] INNER JOIN [UserIndex] ON [UserIndex].[DocumentId] = [Document].[Id] ) yessqlQuery  WHERE Seq BETWEEN 1 AND 10

Exists a simpler way of doing it?

sebastienros:
I merged changes to test on multiple sql server versions (2019 and 2017) https://github.com/sebastienros/yessql/pull/304/files

Please follow this technique to create new test classes (minor changes).

However the image we are using only support 2017 and 2019, we would need an older image to test 2008. Not sure if we can do multiple images/tests in the same appveyor file.

But at least testing that the dialect works on 2019 like you are doing is a minimum. Though it doesn't ensure it works on 2008.

src/YesSql.Provider.SqlServer/SqlServer2008R2Dialect.cs
hishamco: My old Harddisk is corrupted I may need to install SQL Server 2008 R2 and try the unit tests again

jersiovic:
@sebastienros after your merge it seems some of the tests still needs a refactor to use the correct connection string for sql server version we are testing.
I mean at these lines:
https://github.com/sebastienros/yessql/blob/dev/test/YesSql.Tests/SqlServerTests.cs#L53
https://github.com/sebastienros/yessql/blob/dev/test/YesSql.Tests/SqlServerTests.cs#L81
https://github.com/sebastienros/yessql/blob/dev/test/YesSql.Tests/SqlServerTests.cs#L99

jersiovic:
I have run the tests for this PR in a 2008 sql server and I can confirm it doesn't pass any of the two pagination tests.
```
 ShouldPageResults
   Source: CoreTests.cs line 2161
   Duration: 25,7 sec
  Message: 
    Microsoft.Data.SqlClient.SqlException : Invalid column name 'RowNum'.
```
```
 ShouldPageWithoutOrder
   Source: CoreTests.cs line 2242
   Duration: 25,8 sec
  Message: 
    Microsoft.Data.SqlClient.SqlException : Invalid column name 'RowNum'.
```

jersiovic:
I have submited another PR with my fix for pagination, just in case you think it is interesting https://github.com/sebastienros/yessql/pull/305

sebastienros:
Closing as it's old. Still ok to see a new one with passing tests.

PR #65, opened 2017-06-09
Add dialect as optional parameter to the Using extension methods

hishamco:


sebastienros:
I don't think that is sufficient, because it's not just about registering a dialect for a connection type, but having different Stores with the same connection type but different dialects.

For instance in ORchard, one tenant might be on SQL2008 and one on SQL20016.

I think from this PR you can just remove the static dictionary of dialects/connectiontype and put the dialect instance only in the configuration. With the default value like you did.

hishamco:
Is this mean that we need to remove the `SqlDialectFactory`, and add `ISqlDialect` to `IConfigurtion`?

sebastienros:
I think so. Assuming that IConfiguration is accessible anywhere SqlDialectFactory is.

hishamco:
That's what I saw
let me remove the `SqlDialectFactory` and change what is require ..

hishamco:
It seems removing the `SqlDialectFactory` is not easy as we think, there some classes (SchemaBuilder, CommandInterpreterFactory) have a `IDbConnection` object used to retrieve the `ISqlDialect` via `SqlDialectFactory.For`

hishamco:
BTW @sebastienros how you will pass the dialect per tenant AFAIK Orchard installation accepts the provider name & the connection string

sebastienros:
This can be another entry in the setup (Sql Server, Sql Server 2008, Sqlite, ...)

Or as I said, be based on the connection string to pass a custom parameter that the a sql factory could use to return a dialect based on the connection. The connection string is available on the connection, so this could be it. Make a static function instead of an indexer.

hishamco:
So it 'll be part of the connection string, is it something like this 'Server=.;Database=YesSQL;Integrated Security=True;Dialect=SqlServer2008Dialect` after that the dialect will be selected and the parameter will be trimmed from the original connection string

sebastienros:
right, but no need to trim the connection string.

hishamco:
I except the `IDbConnection` implementation will throw an exception if there's a custom param 😄

sebastienros:
I assume it depends on each implementation

hishamco:
I see, let me try it ..

hishamco:
Can we simplify this by using the actual connection string as a key [here](https://github.com/sebastienros/yessql/blob/master/src/YesSql.Provider.SqlServer/SqlServerDbProviderOptionsExtensions.cs#L11-L12) and pass the appropriate dialect after we check the connection string

sebastienros:
This is what I was suggesting, to remove the dictionary and make it a function. You would then register a lambda instead of a type.

hishamco:
Is there any issue if we use the same dictionary? I see `SqlDialectFactory.For` can be used instead

hishamco:
@sebastienros I expect before the keyword is not supported!! Can we use the condition for SQL Server for now to solve the current paging issue and looking back for a proper design to make it generic

sebastienros:
A few comments:
- The dialect should be specified optionally, meaning the defaults should work, and only the custom cases be handled. So we should set Sql2008 only if necessary.
- Please don't enumerate the libraries and types, but have something register dialects with a string, like `Sql2008`, or a lambda that contains the logic to look into the connection string.

If you need I can try to commit something on your PR.

hishamco:
Feel free to add your commits Sebastien

sebastienros:
Obsolete due to new ways of defining the dialect in the configuration, thanks

PR #69, opened 2017-08-21
Netcore20

sebastienros:


hishamco:
@sebastienros why the CI failed after the PR?

sebastienros:
Because the nuget api key expired. I just created a new one and update the script.

PR #76, opened 2017-12-14
Make DefaultIdentifierFactory return private setter

flew2bits:
This change allows the id accessors to set a private "Id" field on an object. This allows for the creation of immutable entities that can be rehydrated from the store since the Id property setter is no longer required to be public.


sebastienros:
Can you add a unit test, or we risk to regress in the future.

flew2bits:
Is the added test sufficient, or do I need to perform some more assertions?

PR #106, opened 2018-05-01
Fix Get to support collections

lunchin:


sebastienros:
Seems like I missed this PR, could you provide a unit test that would exercise the issue?

sebastienros:
I created a unit test and fixed it, thanks.

Issue #104, opened 2018-03-27
Map Reduce Issue


cocoke:
Hi,

I'm trying to use the Map-Reduce system of YesSQL in OrchardCore and i'm facing a problem :
I need to count the number of items inside a document object :

Main document object (one document per minute):
```cs
public class MetricsDocument
{
        public int Id { get; set; }
        public List<Metrics> Metrics = new List<Metrics>();
        public int Serial { get; set; }
        public DateTime Date { get; set; }  
        // Date stops at the minutes
}

public class Metrics : Entity
{
        public DateTime Date { get; set; }
        public string User { get; set; }
        public string Message { get; set; }
        public string Question { get; set; }
        public string Type { get; set; }
        public string Tags { get; set; }
}
```

With this object I build 3 indexes, one is working, but not the two others

First (working) count the number of `Metrics` every minutes:
```cs
public class MetricsIndex : MapIndex
    {
        public DateTime Date { get; set; }
        public int Count { get; set; }
    }

    public class MetricsIndexProvider : IndexProvider<MetricsDocument>
    {
        public override void Describe(DescribeContext<MetricsDocument> context)
        {
            context.For<MetricsIndex>()
                .Map(metric =>
                {
                    return new MetricsIndex
                    {
                        Date = metric.Date,
                        Count = metric.Metrics.Count()
                    };
                });

        }
    }
```

The two other indexes intended to count the number of `Metrics` per hour and per day,
Here is one, the second is pretty similar :
```cs
public class MetricsMessageByDay : ReduceIndex
{
        public string Date { get; set; }
        public int Count { get; set; }
}

public class MetricsMessageByDayIndex : IndexProvider<MetricsDocument>
{
        public override void Describe(DescribeContext<MetricsDocument> context)
        {
            context.For<MetricsMessageByDay, string>()
                .Map(m=> new MetricsMessageByDay
                {
                    Count = m.Metrics.Count(),
                    Date = m.Date.ToString("yyyy/MM/dd")
                })
                .Group(index => index.Date)
                .Reduce(group => new MetricsMessageByDay
                {
                    Date = group.Key,
                    Count = group.Sum(x => x.Count),
                })
                .Delete((index, map) =>
                {
                    return index;
                });
        }
}
```
Sadly this index result count the number of `MetricsDocument` per day and not the number of `Metrics` inside them.

Did I do something wrong or the use of the map reduce system is not the one expected ?

sebastienros:
[...]

PR #111, opened 2018-06-11
Unit Test - map/reduce issue with list of objects

cocoke:
#104


sebastienros:
Shouldn't the Map result return a list of Attachment ByDay in this case? Have to check it's possible though, I had not thought of this case, and it seems common though.

sebastienros:
My solution is valid too, but the bug is still there. Currently working on it.

sebastienros:
Created another PR with the fix

PR #123, opened 2018-09-15
Supporting bigint as primary key

siyamandayubi:
SchemaBuilder.CreateTable(nameof(Sample), table => table
                .Column<long>("SampleId", col => col.Identity().PrimaryKey())
                .Column<string>("Value", col => col.WithLength(200))
            );

generate a table with int primary key instead of bigint primary key.

### DIFF
--- a/src/YesSql.Abstractions/ISqlDialect.cs
+++ b/src/YesSql.Abstractions/ISqlDialect.cs
@@ -20,6 +20,7 @@ namespace YesSql
         bool HasDataTypeInIdentityColumn { get; }
         bool SupportsIdentityColumns { get; }
         string IdentityColumnString { get; }
+        string BigIntIdentityColumnString { get; }
         string IdentitySelectString { get; }
         string GetTypeName(DbType dbType, int? length, byte precision, byte scale);
         string GetSqlValue(object value);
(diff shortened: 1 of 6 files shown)


sebastienros:
Can you elaborate for the needs of using `bigint` for ids? Is it that you are afraid of being limited over time?
Shouldn't it also update the type for the Identity table that contains the ranges of ids for inserts?

hishamco:
> Is it that you are afraid of being limited over time?

Many of us even I 😄 was worried little bit to reach the limit over the time

siyamandayubi:
@sebastienros Maybe I shouldn't be worry about being limited over time, but it is the only reason. By updating the Identity Column, you mean a scenario that we have an existing table with Identity int32 and then we have AlterTable command that change the Identity to 64? I didn't check that, I will try it and try to find a way to avoid it

sebastienros:
What I mean is that there are other tables in the system that use `int` and should be converted to `bigint` in that case, otherwise you are only fixing part of the problem. Also there is some reflection code around ids, that might need to be adapted. The best way would be to add a unit test that seeds the ids at a number above the `int` limit and creates some data.

A concern could be perf also, I believe the db could be slower with `bigint`, but I am not sure how much. Same, a benchmark would show it.

siyamandayubi:
@sebastienros  Now I got what you mean. Actually, the usecase for supporting bigint for primary key is not for Index tables or main Document table.  Alongside Document and its associated index tables, there are some other tables that would contain much higher volume of records and I want to manage them through the same Migration files and SchemaBuilder.

Issue #122, opened 2018-09-14
Sqlite quoting


sebastienros:
c.f. https://www.sqlite.org/lang_keywords.html

Currently the dialect is using `"` to quote identifiers. As documented, if the identifier doesn't exist and the context of the query allows for strings, then it will be interpreted as a string. Hence, doing `SELECT "not a column"` will return it like it was a string and not a column.

The issue is that in case of a typo in the column name, there won't be an error message and the string itself will be used.

Solutions are
- Use `[ ]` as it's supported for SQL Server compatibility
- User back ticks as it's supported for MySQL compatibility

PR #127, opened 2018-09-18
Use square brackets instead of double quotation

hishamco:
Address issue #122


sebastienros:
I am ok with removing the escaping as having brackets in a table name is far from obvious.

Issue #121, opened 2018-09-14
Use COLLATE NOCASE by default for SQLite


sebastienros:
Text queries are case sensitive by default. Add it the creation scripts, include the index as they need to match.

hishamco:
As `NOCASE` is a collation for case-insensitive comparing. But it doesn’t support Unicode only ASCII

hishamco:
FYI we can use `COLLATE NOCASE` in the column creation

sebastienros:
@hishamco yes, this is what I had in mind. It will require to add some test for case insensitive matches, and also for unicode chars which I think we don't have yet.

hishamco:
Hope to get on this soon

PR #128, opened 2018-09-24
Add nocase collation to SqLite

hishamco:
Address issue #121


hishamco:
@sebastienros this is WIP, but i'm wondering if there's a better way to override the `CreateColumnCommand` instead of repeating `ICreateTableCommand` implementations

sebastienros:
Can't you just add `collate nocase` to the `TEXT` value in the Sqlitedialect ?

sebastienros:
Or just change the dialect interface such that it can have an extra piece of text added in DDL scripts

sebastienros:
Also, don't forget to add a test case that would order a list of elements based on a string property, and has a mix of upper case and lower case values for instance `[ 'Add', 'added', 'Addition', 'Be', 'behold']`

hishamco:
> Can't you just add collate nocase to the TEXT value in the Sqlitedialect ?

I was thinking of such thing before I push the PR 😄, let me check if this is doable or shall I change the dialect interface 

> Also, don't forget to add a test case that would order a list of elements based on a string property

Of course .. that's while I told you it's still WIP

src/YesSql.Core/Sql/BaseComandInterpreter.cs
```diff
@@ -246,23 +246,27 @@ public virtual IEnumerable<string> Run(IDropForeignKeyCommand command)
...
 
             // append identity if handled
             if (command.IsIdentity && _dialect.SupportsIdentityColumns)
             {
                 builder.Append(Space).Append(_dialect.IdentityColumnString);
+                ignoreCollation = true;
+            }
+
+            // collation
+            if (!ignoreCollation)
```
hishamco: I knew this is weird little bit, because you inject the `Integer` data type on the `IdentityColumnString`, IMHO we should rid off the data type from `IdentityColumnString`, so this below line will be applied for all the columns in the table
```csharp
builder.Append(_dialect.GetTypeName(command.DbType, command.Length, command.Precision, command.Scale));
```

hishamco:
@sebastienros I used a compiled queries for the unit tests, if this is fine, please do a final review

sebastienros:
The whole PR is a big HACK that happens to work.

Have you at least tried to __just__ change `"TEXT"` to `"TEXT COLLATE NOCASE"` in the list of types? And that's it, can't be simpler than that for the exact same features.

hishamco:
> The whole PR is a big HACK that happens to work.

As I mentioned above on the comment I made the HACK because you stick the `integer` with `IdentityColumnString`, please refer to my comment [here](https://github.com/sebastienros/yessql/pull/128#discussion_r220078666)

> Have you at least tried to just change "TEXT" to "TEXT COLLATE NOCASE"

I think it's a bad idea 😞, if we rid off the `integer` part there will be no HACK, because `ColumnCollation`  will have `COLLATE NOCASE` that come from the `SqLiteDialect`

sebastienros:
I don't understand what you call "the integer part "

hishamco:
If you look closely to [this](https://github.com/sebastienros/yessql/blob/dev/src/YesSql.Provider.Sqlite/SqliteDialect.cs#L51) line, you will see that the `integer` attached here which is should be come from `GetTypeName()` if this is fixed the code will become
```csharp
if(ColumnCollation != String.Empty)
{
     builder.Append(Space).Append(_dialect.ColumnCollation);
}
```
so no HACK is needed

sebastienros:
Thank you for your effort but 
- you are inventing features this PR is not supposed to solve
- you are mutating the dialect instance while its types are resolved (`ColumnCollation` changes on the fly without being reset)

If you think there is a problem with how the identity column works, file an issue showing what scenario you can't achieve. Because I still don't know what problem you are trying to solve. It might just be you are not at ease with how things work in YesSql, but I don't want to add complexity that is not adding any value for the end user.

hishamco:
It's fine .. if I knew that breaking rules is allowed the PR is nothing but adding a `COLLATE NOCASE` to a TEXT data type 😄 

But again i'm not wrong this doesn't solve a unicode ordering

PR #129, opened 2018-09-25
Using case insensitive collation in Sqlite

sebastienros:


sebastienros:
@hishamco What is wrong with these changes?

If I remove the change, the new unit test fails. If I apply the change, all unit tests still pass and the new one too. Can you add a unit test that breaks "the integer part" ?

PR #130, opened 2018-10-09
Upgrade Dependencies

agriffard:
MySql, PostgreSql


sebastienros:
We don't have to change these in yessql, we can do it directly in OC by adding a reference.

agriffard:
The build is still failing but I don't know how to fix it. 
I may be should have let the references as they were in YesSql.Samples.Performance.csproj.

sebastienros:
It's working for me.

Issue #133, opened 2018-10-15
ISession.Query<T> is reported as blocking on an async thread


johnrom:
Running the following command is reported as blocking on an async thread:

```
var result = await session.Query<T>().ListAsync();
```

The log, reported via [Ben.BlockingDetector](https://github.com/benaadams/Ben.BlockingDetector):

```
Ben.Diagnostics.BlockingMonitor: Warning: Blocking method has been invoked and blocked, this can lead to threadpool starvation.
   at System.Threading.WaitHandle.WaitAny(WaitHandle[] waitHandles, Int32 millisecondsTimeout, Boolean exitContext)
   at System.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, UInt32 waitForMultipleObjectsTimeout, Boolean allowCreate, Boolean onlyOneCheckConnection, DbConnectionOptions userOptions, DbConnectionInternal& connection)
   at System.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, TaskCompletionSource`1 retry, DbConnectionOptions userOptions, DbConnectionInternal& connection)
   at System.Data.ProviderBase.DbConnectionFactory.TryGetConnection(DbConnection owningConnection, TaskCompletionSource`1 retry, DbConnectionOptions userOptions, DbConnectionInternal oldConnection, DbConnectionInternal& connection)
   at System.Data.ProviderBase.DbConnectionInternal.TryOpenConnectionInternal(DbConnection outerConnection, DbConnectionFactory connectionFactory, TaskCompletionSource`1 retry, DbConnectionOptions userOptions)
   at System.Data.ProviderBase.DbConnectionClosed.TryOpenConnection(DbConnection outerConnection, DbConnectionFactory connectionFactory, TaskCompletionSource`1 retry, DbConnectionOptions userOptions)
   at System.Data.SqlClient.SqlConnection.TryOpen(TaskCompletionSource`1 retry)
   at System.Data.SqlClient.SqlConnection.Open()
   at YesSql.Session.Demand()
   at YesSql.Session.Query()
   at YesSql.QueryExtensions.Query[T](ISession session)
   at MyProject.Modules.Global.Services.TestimonialsService.<<Get>b__5_0>d.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
[...]

PR #134, opened 2018-10-15
Using DbConnection.OpenAsync

sebastienros:


sebastienros:
Fixes https://github.com/sebastienros/yessql/issues/133

PR #138, opened 2018-11-09
Return documents in order of ids that were provided

jameshawkins:
Related to https://github.com/OrchardCMS/OrchardCore/issues/2651

### DIFF
--- a/test/YesSql.Tests/CoreTests.cs
+++ b/test/YesSql.Tests/CoreTests.cs
@@ -2267,7 +2267,7 @@ namespace YesSql.Tests
 
             using (var session = _store.CreateSession())
             {
-                var circle = await session.GetAsync<Circle>(circleId);
+                var circle = await session.GetAsync<object>(circleId);
 
                 Assert.NotNull(circle);
                 Assert.Equal(typeof(Circle), circle.GetType());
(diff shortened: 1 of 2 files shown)


test/YesSql.Tests/CoreTests.cs
```diff
@@ -2301,6 +2301,39 @@ public async Task ShouldGetDynamicById()
...
+                    session.Save(circle);
+                    circleIds.Add(circle.Id);
+                }
+
+                await session.CommitAsync();
+            }
+
+            using (var session = _store.CreateSession())
+            {
+                circleIds.Reverse();
```
sebastienros: The test is wrong, I don't think `Reverse()` mutates the callee. It returns a new collection instead. So your test is still testing the same initial order.

```suggestion
                circleIds = circleIds.Reverse();
```

sebastienros: Add `.ToList()` to that

jameshawkins: it mutates the list: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.reverse?view=netframework-4.7.2

jameshawkins: i was surprised too :)

sebastienros:
Now just wait for the build to finish and update the version on Orchard in a PR

jameshawkins:
Orchard PR submitted: https://github.com/OrchardCMS/OrchardCore/pull/2657

Issue #142, opened 2019-01-08
Int64 document identifiers


CoskunSunali:
What would it take to be able to create the Document table with a primary key of `Int64` value (bigint)?

That is for products expected to have more than 2.1 billion documents (including those deleted) which is not unlikely.

sebastienros:
I assume we should just change all the code to use this ... Do you know it it's supported in all dbs and the impact on perf it would have? I'd like to know if it would be necessary to support both.

CoskunSunali:
As far as I know, all providers that YesSql support so far (mysql, sql server, postgresql and sqlite) support `bigint` column type.

Event though I don't have real life benchmarks, I am almost certain that the performance will be effected negatively in a scale of some milliseconds. The storage size will increase after all, indexes will be bigger, disk IO will probably cost more time to process.

However, we shall not forget that the performance lose doesn't matter as soon as the int32 primary keys start overflowing and break the application.

There are some disk size comparisons for MySQL at http://ronaldbradford.com/blog/bigint-v-int-is-there-a-big-deal-2008-07-18/

These guys find similar results for int/bigint on PostgreSQL at https://stackoverflow.com/questions/38053596/benchmark-bigint-vs-int-on-postgresql

There is a very details comparison for MSSQL at https://www.mssqltips.com/sqlservertip/5115/compare-sql-server-table-join-performance-for-int-vs-bigint-vs-guid-data-types/

I don't have any links for SQLite but as far as I know SQLite automatically decides whether it should use int or bigint based on the magnitude of the value.

IanYates:
Related PR here
https://github.com/sebastienros/yessql/pull/123

sebastienros:
@IanYates I knew there was something like this, couldn't find it in the issues ;) Thanks

CoskunSunali:
Thanks @IanYates. It seems that there is some more work to do there. E.g. index tables should have `long` identifiers as well as several methods should accept long instead of int too. @sebastienros already mentions them [here](https://github.com/sebastienros/yessql/pull/123#issuecomment-422435816).

Do you think you can complete it any time soon?

I know it is easy to change the things when you want to support either int or long but gets complicated when you want to support both.

sebastienros:
Taking another look at the PR, I see two different features are requested.
- the fact that CreateColumn should allow to create identity columns that are bigint
- have all Document and Index table use bigint by default (even if the classes use int only)

The PR implements the first case. The current issue asks for the second.

[...]

PR #143, opened 2019-01-09
Change primary key types from int to bigint.

CoskunSunali:
Related to https://github.com/sebastienros/yessql/issues/142

### DIFF
--- a/src/YesSql.Abstractions/Document.cs
+++ b/src/YesSql.Abstractions/Document.cs
@@ -5,7 +5,7 @@ namespace YesSql
         /// <summary>
         /// The unique identifier of the document in the database.
         /// </summary>
-        public int Id { get; set; }
+        public long Id { get; set; }
 
         /// <summary>
         /// The type of the document.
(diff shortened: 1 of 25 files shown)


siyamandayubi:
I think an ideal solution is supporting both of int32 and int64 keys. For example, by default, everything can be Int32 as currently it is and if someone wants to have Int64 keys, he has to use Store<long> and Session<long>.

> public class Store< T> : IStore< T> where T: IConvertible

and by default

> public class Store : Store< int>


and in this way, we have refactor the necessary subclasses, it would be a big refactoring, but doing it would not be so difficult as for many classes, it would be just adding generic to the class and having a default one with int32. @sebastienros what do you think?

CoskunSunali:
> Sorry, I have done quite many breaking changes. If you don't have time to merge these, I'll at doing it by myself.

@sebastienros Unfortunately don't have the time to merge :(

jersiovic:
So, is it planned to merge this PR in the short term?
Thank you

Issue #147, opened 2019-01-25
Document types cannot be shared between assemblies


wedge206:
I have two different apps sharing the same DB.  I've got them sharing the same models with the same namespaces using a Shared Project.
Each app cannot read data written by the other app, because the SimplifiedTypeName() extension includes the assembly name.  So the data from each assembly is ignored by the other.
I don't see any way around this currently.

Right now the Document Type value is hardcoded as:
`String.Concat(type.FullName, ", ", typeInfo.Assembly.GetName().Name));`

That's fine as a default Type name.  
However, I feel we should be able to specify our own Type names.  We should be able to override the SimplifiedName with an attribute on the model class.  And/Or we should be able to override the assembly name with a property on the Store. Configuration.
Something like that should solve for my use case.  Or let me know if there's already some way to do what I need, and I'm just missing it?

If you like the idea, I can submit a PR to update the SimplifiedTypeName() method the way I describe.

Something like this:
```
    public class SimplifiedTypeName : Attribute
    {
        public string Name { get; private set; }

        public SimplifiedTypeName(string name)
        {
            Name = name;
        }
    }

[SimplifiedTypeName("Custom Simplified Name")]
    public class MyModel
{
....
}
```

Then update the SimplifiedTypeName() extension to use the custom value if it exists, and if not then use the default.

sebastienros:
I suggest that a dictionary of `<Type, string>` is added to the store, and we can cache the names there.
Then if no mapping is found we would 
1- look for this attribute you are suggesting
2- build a simplified name as of today

This configuration would be public, and this would allow an external code to also configure it.

sebastienros:
Please implement this as a PR if you can

PR #148, opened 2019-01-25
Implemented KnownTypeNames collection and custom SimplifiedTypeName attribute

wedge206:
Not sure if that's exactly what you had in mind for the dictionary on the store.  Let me know if you have any suggestions/changes.

#147


wedge206:
Had to move it to Abstractions.  That's the only place that made sense that didn't create circular dependencies.

src/YesSql.Abstractions/TypeService.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                var typeInfo = t.GetTypeInfo();
+                if (IsAnonymousType(typeInfo))
+                {
+                    return "dynamic";
+                }
+
+                var customName = typeInfo.GetCustomAttribute<SimplifiedTypeName>();
+
+                return typeNames.GetOrAdd(t, String.IsNullOrEmpty(customName?.Name) ? String.Concat(t.FullName, ", ", typeInfo.Assembly.GetName().Name) : customName.Name);
```
sebastienros: The "updated" value should be a lambda or it will be computed on every call.

src/YesSql.Abstractions/TypeService.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public class TypeService : ITypeService
+    {
+        private readonly ConcurrentDictionary<Type, string> typeNames = new ConcurrentDictionary<Type, string>();
+
+        public string this[Type t]
+        {
+            get
+            {
+                var typeInfo = t.GetTypeInfo();
```
sebastienros: This and the `customName` generation should be in the concurrentdictionary lambda

src/YesSql.Abstractions/TypeService.cs
```diff
@@ -0,0 +1,42 @@
+using System;
```
sebastienros: Last thing, move this class out of Abstractions.

PR #163, opened 2019-02-23
fix bug in saving data

siyamandayubi:
If an exception happens during saving the data, the _flashing flag stays
true and while there is a session-pooling (and the last session has _flashing=true flag), the next operations will not save data.

### How to reproduce

- Save a document that raises an exception, (for example, by setting too long string for value of one of the indexes)

- try to create a new Session from the same  Store and save another object. (Assuming the Store use the previous Session of the Session-Pool, the second object will be never saved as the _flashing flag stayed true

### DIFF
--- a/src/YesSql.Core/Session.cs
+++ b/src/YesSql.Core/Session.cs
@@ -476,49 +476,58 @@ namespace YesSql
                 return;
             }
 
-            _flushing = true;
-
-            // we only check if the session is disposed if 
-            // there are no commands to commit.
-
-            CheckDisposed();
-
-            // saving all updated entities
-            foreach (var obj in _updated)
+            try
             {
-                if (!_deleted.Contains(obj))
+                _flushing = true;
+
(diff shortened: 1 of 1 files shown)


siyamandayubi:
@sebastienros  did you forget this one ;-)

src/YesSql.Core/Session.cs
```diff
@@ -476,49 +476,58 @@ public async Task FlushAsync()
...
+                _flushing = false;
+                throw;
             }
 
-            _updated.Clear();
-            _saved.Clear();
-            _deleted.Clear();
-            _commands.Clear();
-            _maps.Clear();
             _flushing = false;
```
hishamco: Can we use `finally` block instead of setting _flushing twice
```csharp
finally
{
    _flushing = false;
}
```

siyamandayubi: @hishamco  yep, I changed it

PR #168, opened 2019-03-06
fix mysql foreign key issue

siyamandayubi:
Unlike SQl Server, by default Mysql sets the foregin key as Restricted which causes a problem,
if you want to delete a document

### DIFF
--- a/src/YesSql.Provider.MySql/MySqlDialect.cs
+++ b/src/YesSql.Provider.MySql/MySqlDialect.cs
@@ -98,24 +98,9 @@ namespace YesSql.Provider.MySql
 
         public override string GetAddForeignKeyConstraintString(string name, string[] srcColumns, string destTable, string[] destColumns, bool primaryKey)
         {
-            var res = new StringBuilder(200);
+            string sql = base.GetAddForeignKeyConstraintString(name, srcColumns, destTable, destColumns, primaryKey);
 
-            if (SupportsForeignKeyConstraintInAlterTable)
-                res.Append(" add");
-
-            res.Append(" constraint ")
-                .Append(name)
-                .Append(" foreign key (")
-                .Append(String.Join(", ", srcColumns))
-                .Append(") references ")
-                .Append(destTable);
-
-            if (!primaryKey)
-            {
-                res.Append(" (")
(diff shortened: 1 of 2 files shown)


sebastienros:
Could you add a unit test that would fail without the fix?

siyamandayubi:
Yep, a UnitTest has been added.

test/YesSql.Tests/MySqlTests.cs
```diff
@@ -37,5 +41,36 @@ protected override void OnCleanDatabase(SchemaBuilder builder, DbTransaction tra
             }
             catch { }
         }
+
+        [Fact]
+        public async Task ForeignKeyOfIndexesMustBe_DeleteCascated()
+        {
+            var configuration = CreateConfiguration();
+            _store = StoreFactory.CreateAsync(configuration).GetAwaiter().GetResult();
```
sebastienros: await

test/YesSql.Tests/MySqlTests.cs
```diff
@@ -37,5 +41,36 @@ protected override void OnCleanDatabase(SchemaBuilder builder, DbTransaction tra
...
+                Lastname = "Gates"
+            };
+
+            using (var session = _store.CreateSession())
+            {
+                session.Save(bill);
+            }
+
+            // second store, don't register the index
+            _store = StoreFactory.CreateAsync(configuration).GetAwaiter().GetResult();
```
sebastienros: await

src/YesSql.Provider.MySql/MySqlDialect.cs
```diff
@@ -96,6 +96,33 @@ public override string GetDropForeignKeyConstraintString(string name)
             return " drop foreign key " + name;
         }
 
+        public override string GetAddForeignKeyConstraintString(string name, string[] srcColumns, string destTable, string[] destColumns, bool primaryKey)
+        {
+            var res = new StringBuilder(200);
```
sebastienros: call `base.GetAdd...`

siyamandayubi:
@sebastienros  I applied the changes we talked about it.

siyamandayubi:
@sebastienros  I reverted the password

Issue #175, opened 2019-03-26
Signed Assemblies?


wedge206:
Some build/deploy systems require assemblies and their dependencies to be signed and require a strong name.
Can we add a Strong Name to each of the Yessql assemblies?

sebastienros:
Yes. Do you want to create a PR?

wedge206:
Sure.
I'll probably do secondary signed projects, instead of signing the existing projects.  Similar to how Dapper does it.  That way the signed yessql projects can reference the signed Dapper packages.  This way there is no need to change references on the existing projects.

hishamco:
This should be closed

PR #176, opened 2019-04-06
Added signed packages

wedge206:


wedge206:
#175

wedge206:
I'm not sure if any scripts need to be updated to get these to build and push to nuget.

sebastienros:
I'd rather ship only a single version that is strong named. Also pointed here: https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming

wedge206:
Interesting.  Releasing two packages (strong named, and not strong named) is such a common practice, I didn't realize it was against the official guidelines.
I'll update this PR.

sebastienros:
I also do a single package for Jint and Esprima. It has always worked for me.

Thanks

wedge206:
Updated

sebastienros:
I didn't understand why I was still seeing all previous commits in the PR, and thought squashing would resolve it. You actually targeted master and didn't realize. I canceled the build and force-pushed to revert the commit. I will try to re-apply it on dev.

wedge206:
Yeah, I noticed when I got the “merged to master” notification

Issue #95, opened 2018-01-29
Custom logger


sebastienros:
Allow a store to be passed a custom `ILogger` to use to log queries. This should use the standard dotnet core logger interface. A `NullLogger` can be used by default.

hishamco:
Sounds good

PR #179, opened 2019-05-08
Implements Microsoft.Extensions.Logger

wedge206:
Removes the YesSql.Logger types.
Added UseLogger() Configuration extension.
Replaced LogSql() with LogTrace()

Also updated some dependencies.


wedge206:
#95

sebastienros:
LGTM

Thanks a lot. IF it's ready let me know and I'll merge it when the tests have passed.

wedge206:
It *should* be working fine.  I haven't tested it with AppInsights yet.  That's my primary use-case, so I should probably make sure it's actually working there before merging.  I'll let you know once I've done some more tests.

wedge206:
@sebastienros Yup.  Confirmed it's working nicely.

wedge206:
A thought I just had is to add the ability to configure which log level SQL statements will be sent at.  I hardcoded it to Trace level.  But it would be easy enough to put the LogSql() method back, and have it send to Trace/Debug/Info/Any level based on configuration.
I'll can add that later if needed.  Just wanted to make a note of it here.

sebastienros:
I think it's good, we just need to configure these namespaces for TRACE if necessary

PR #180, opened 2019-05-14
Added SQL Query debugging

wedge206:
Added LogDebug() statements for SQL queries.


sebastienros:
Let me know when you are done with it. Look good.

wedge206:
It's done.  As far as I can tell all SQL calls are now being logged.  Unless you know any that I missed.

wedge206:
Right now all "system" sql calls log as Trace, and user queries log as Debug.  Seems logical to me.

Issue #183, opened 2019-05-30
ReIndexing can fail when using custom type names


wedge206:
When using custom type names and attempting to reindex the db in a manner like this:

`var entityList = await session.Query().Any().ListAsync();
foreach (var entity in entityList)
{
    session.Save(entity);
}`

The re-index does not work, because all objects are retrieved as JsonObjects, and then its unable to map the type to its indexproviders.  Seems like the fix will have to be around how the IContentSerializer is called.
I'll work on a fix, but its low-pri for now since there is a decent workaround

Workaround:
Retrieve documents explicitly by type.  eg:
`var entityList = await session.Query<T>().ListAsync();
foreach (var entity in entityList)
{
    session.Save(entity);
}`

sebastienros:
There is a unit test that ensure `Any()` resolves the correct type. And I just tried this new one:

```
[Fact]
        public async Task ShouldResolveManyTypes()
        {
            using (var session = _store.CreateSession())
            {
                var bill = new Person
                {
                    Firstname = "Bill",
                    Lastname = "Gates"
                };

                var lion = new Animal
                {
                    Name = "Lion"
                };

                session.Save(bill);
                session.Save(lion);
            }

            using (var session = _store.CreateSession())
            {
                var all = await session.Query().Any().ListAsync();

                Assert.Equal(2, all.Count());
                Assert.Contains(all, x => x is Person);
                Assert.Contains(all, x => x is Animal);
            }
        }
```

wedge206:
The test probably isn’t testing using a custom type name.  That test will need to be added too.

sebastienros:
I can see a difference with you code, I will try to call `Save`. However if the code you shared `Save()` would be no-op because nothing would have changed. We might need a custom overload to force it, in case you want the indices to be computed again.

wedge206:
If you add this in the CreateTables():
`_store.TypeNames[typeof(Person)] = "People";`

Then the ShouldResolveManyTypes() test fails.

sebastienros:
"ReIndexing can fail when using __custom type names__"

Now I understand what you mean

wedge206:
I think I have a fix, updating the TypeService.  It requires a reverse lookup from the name to the type.

PR #185, opened 2019-05-31
Added ReverseLookup to TypeService

wedge206:
This is required to fix a bug when queries do not specify a type.

I don't know if the method should call FirstOrDefault() or SingleOrDefault().
I used FirstOrDefault for safety.  But I feel Single is more correct.  There should never be more than one type mapped to the same value.

#183

### DIFF
--- a/src/YesSql.Abstractions/ITypeService.cs
+++ b/src/YesSql.Abstractions/ITypeService.cs
@@ -1,9 +1,16 @@
 using System;
+using System.Collections.Generic;
 
 namespace YesSql
 {
     public interface ITypeService
     {
-         string this[Type t] { get; set; }
+        string this[Type t] { get; set; }
+
+        IEnumerable<Type> Keys { get; }
+
+        IEnumerable<string> Values { get; }
+
+        Type ReverseLookup(string value);
     }
 }
--- a/src/YesSql.Abstractions/ITypeService.cs
+++ b/src/YesSql.Abstractions/ITypeService.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 
 namespace YesSql
 {
@@ -7,10 +6,6 @@ namespace YesSql
     {
         string this[Type t] { get; set; }
 
-        IEnumerable<Type> Keys { get; }
-
-        IEnumerable<string> Values { get; }
-
-        Type ReverseLookup(string value);
+        Type this[string s] { get; }
     }
 }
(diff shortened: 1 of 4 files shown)


wedge206:
Note: the Keys/Values properties are not part of the fix.  But I found myself wanting those properties when I was searching for a workaround.  So they seemed useful enough to add.

src/YesSql.Abstractions/ITypeService.cs
```diff
@@ -1,9 +1,16 @@
 using System;
+using System.Collections.Generic;
 
 namespace YesSql
 {
     public interface ITypeService
     {
-         string this[Type t] { get; set; }
+        string this[Type t] { get; set; }
```
sebastienros: Maybe only add one more indexer. `Type this[string t]` and remove the other properties from the interface.

src/YesSql.Core/Services/TypeService.cs
```diff
@@ -8,6 +10,15 @@ public class TypeService : ITypeService
     {
         private readonly ConcurrentDictionary<Type, string> typeNames = new ConcurrentDictionary<Type, string>();
 
+        public IEnumerable<Type> Keys { get { return typeNames.Keys; } }
+
+        public IEnumerable<string> Values { get { return typeNames.Values; } }
+
+        public Type ReverseLookup(string value)
+        {
+            return typeNames.FirstOrDefault(t => t.Value == value).Key;
```
sebastienros: You actually need two `ConcurrentDictionary`. And when a vale is set on the indexers, just set the two dictionaries.

wedge206: Two dictionaries makes sense.  That would guarantee uniqueness in both directions.

src/YesSql.Abstractions/ITypeService.cs
wedge206: That doesn't feel right, I think that would add some confusion.  With a single indexer it should be obvious what the intended usage is.  But with two separate indexers, the behavior will not obvious just by looking at it.

sebastienros: Convert it to method names if you prefer

wedge206: Looked at it a few different ways.  I think two indexers makes sense as long as only one of them allows setting

msrcgit:
Is this one okay to be approved?

wedge206:
Is this one okay to be approved?

sebastienros:
Thanks, had forgotten about it

Issue #184, opened 2019-05-31
DropIndex fails on sqlite


deanmarcussen:
The `AlterTable.DropIndex` command fails on Sqlite

Error is `Error : "near "ON": syntax error`

Sqlite doesn't support ON as part of the `DROP INDEX` command

Correct syntax (for sqlite) seems to be
`DROP INDEX "IDX_AliasPartIndex_Alias";`

In `BaseCommandInterpreter` it's part of the general command
```
        public virtual void Run(StringBuilder builder, IDropIndexCommand command)
        {
            builder.AppendFormat("drop index {0} ON {1}",
                _dialect.QuoteForColumnName(command.IndexName),
                _dialect.QuoteForTableName(command.Name));
        }
```

PR #186, opened 2019-06-05
Fixes drop index sql dialect

deanmarcussen:
Fixes https://github.com/sebastienros/yessql/issues/184

I've tested this on MS SQL and SQLite and works well.

Research says that MySQL supports `ON table_name` syntax when dropping indexes, but PostgreSql does not, so have coded accordingly.

May not actually need to use this feature for the issue I was trying to fix (change length of column), but as many database providers are involved and can behave differently , it seemed sensible to be able to drop the index, change the column length, and recreate the index

### DIFF
--- a/src/YesSql.Core/Sql/BaseComandInterpreter.cs
+++ b/src/YesSql.Core/Sql/BaseComandInterpreter.cs
@@ -203,9 +203,7 @@ namespace YesSql.Sql
 
         public virtual void Run(StringBuilder builder, IDropIndexCommand command)
         {
-            builder.AppendFormat("drop index {0} ON {1}",
-                _dialect.QuoteForColumnName(command.IndexName),
-                _dialect.QuoteForTableName(command.Name));
+            builder.Append(_dialect.GetDropIndexString(command.IndexName, command.Name));
         }
 
         public virtual IEnumerable<string> Run(ISqlStatementCommand command)
--- a/src/YesSql.Provider.Common/BaseDialect.cs
+++ b/src/YesSql.Provider.Common/BaseDialect.cs
@@ -89,6 +89,8 @@ namespace YesSql.Provider
         }
 
         public virtual bool SupportsIfExistsBeforeTableName => false;
+        public virtual bool SupportsIfExistsBeforeIndexName => true;
+        public virtual bool SupportsTableNameAfterIndexName => true;
         public virtual string CascadeConstraintsString => String.Empty;
         public virtual bool SupportsIfExistsAfterTableName => false;
         public virtual string GetDropTableString(string name)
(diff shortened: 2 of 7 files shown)


deanmarcussen:
Now tested on PostgreSql too...

I didn't write any unit tests for this, but I see how the unit tests are built now. But I didn't really see a set of tests around basic command syntax like DropIndex or AlterTable, so let me know if you'd like me to write a test for it?

src/YesSql.Provider.Common/BaseDialect.cs
```diff
@@ -108,6 +110,24 @@ public virtual string GetDropTableString(string name)
...
+            if (SupportsIfExistsBeforeIndexName)
+            {
+                sb.Append("if exists ");
+            }
+
+            sb.Append(QuoteForColumnName(indexName));
+
+            if (SupportsTableNameAfterIndexName)
+            {
+                sb.AppendFormat(" on {0}", QuoteForTableName(tableName));
```
sebastienros:
```suggestion
                sb.Append(" on ").Append(QuoteForTableName(tableName));
```

src/YesSql.Provider.Common/BaseDialect.cs
```diff
@@ -89,6 +89,8 @@ public virtual string GetDropForeignKeyConstraintString(string name)
         }
 
         public virtual bool SupportsIfExistsBeforeTableName => false;
+        public virtual bool SupportsIfExistsBeforeIndexName => true;
```
sebastienros: I would assume that the `GetDropIndexString` is all you need, just update it for each dialect that has a different behavior than the most common way.

deanmarcussen: All 4 dialects are different in one way or another.
Shall I just make an abstract `GetDropIndexString` method and implement for each dialect? 

Just trying to follow the existing patterns...

sebastienros: Yes, I would prefer a single method in this case, an no custom flags. I don't remember the history of the other flag you took as an example. Doesn't look like something I would have done ;)

deanmarcussen: Cool, yeah I couldn't figure them out for ages either

src/YesSql.Core/Sql/BaseComandInterpreter.cs
```diff
@@ -203,9 +203,7 @@ public virtual void Run(StringBuilder builder, IAddIndexCommand command)
 
         public virtual void Run(StringBuilder builder, IDropIndexCommand command)
         {
-            builder.AppendFormat("drop index {0} ON {1}",
-                _dialect.QuoteForColumnName(command.IndexName),
-                _dialect.QuoteForTableName(command.Name));
+            builder.Append(_dialect.GetDropIndexString(command.IndexName, command.Name));
```
sebastienros: Ideally we could have created a method name `DropIndex` that would have accepted a `StringBuilder` (like `Page()`) and reuse this one in it. But that's ok, we don't drop indexes on all calls.

sebastienros:
Waiting for the build to pass before merging.

src/YesSql.Core/Sql/BaseComandInterpreter.cs
deanmarcussen: Ah, ok. To reduce allocations of StringBuilder I assume, makes sense :)

PR #191, opened 2019-07-30
MS SQL Add brackets around TOP value to provide support for parameters

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/3986

Seemed easiest to always put the brackets around the `TOP value` as SQL Server seems happy if its `TOP (2)` or `TOP (@parameter)` but always fails with `TOP @parameter`

Will break the OC tests :)


sebastienros:
Also all the examples in the doc are using parens. https://docs.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql?view=sql-server-2017

Issue #202, opened 2019-09-25
Distinct Documents from Index


wedge206:
The Scenario:

I have a document type that contains a List.
The individual rows from the list are being indexed, so the document can be searched based on the contents of its child list

The Problem:
Some search results will return duplicate documents.  This makes sense because the documents are duplicated in the index.
But what we want/expect is to just return a single instance of a Document, even if it exists multiple times in the Index.

We cannot easily do a Distinct() after the query has executed, because that will break the page sizing.

I don't see any solution or workaround to this problem at the moment.
I feel like a Distinct command or something like it on the binding from the Index to the Document table should be the default behavior.  But that could break other peoples expectation.

sebastienros:
What is the query you are using? 

I could imagine that currently there is an INNER JOIN with the Document table and would return many results. However even if many documents are returned, they should all be the same instance in terms of C# object (can you check)?
Then yes we can probably do a distinct automatically for some queries that join to the document table.

wedge206:
Yes, as far as I can tell it is returning the same instance twice.

Difficult to describe the query.  It's big and it gets assembled based on multiple criteria.

sebastienros:
Maybe write a unit test in a PR that will fail. I assume we already have the domain models necessary to replicate.

wedge206:
Already started doing exactly that

PR #203, opened 2019-09-25
Distinct Document support

wedge206:
#202

### DIFF
--- a/src/YesSql.Core/Services/DefaultQuery.cs
+++ b/src/YesSql.Core/Services/DefaultQuery.cs
@@ -784,7 +784,15 @@ namespace YesSql.Services
             var transaction = await _session.DemandAsync();
 
             var localBuilder = _queryState._sqlBuilder.Clone();
-            localBuilder.Selector("count(*)");
+
+            if (localBuilder.HasJoin)
+            {
+                localBuilder.Selector($"count(distinct {_queryState._sqlBuilder.FormatColumn(_queryState._documentTable, "id")})");
+            }
+            else
+            {
+                localBuilder.Selector($"count(*)");
+            }
 
             // Clear paging and order when counting 
             localBuilder.ClearOrder();
@@ -1006,6 +1014,7 @@ namespace YesSql.Services
                         }
 
                         _query._queryState._sqlBuilder.Selector(_query._queryState._sqlBuilder.FormatColumn(_query._queryState._documentTable, "*"));
+                        _query._queryState._sqlBuilder.Distinct();
                         var sql = _query._queryState._sqlBuilder.ToSqlString();
                         var key = new WorkerQueryKey(sql, _query._queryState._sqlBuilder.Parameters);
                         var documents = await _query._session._store.ProduceAsync(key, (args) =>
--- a/src/YesSql.Core/Services/DefaultQuery.cs
+++ b/src/YesSql.Core/Services/DefaultQuery.cs
@@ -791,7 +791,7 @@ namespace YesSql.Services
             }
             else
             {
-                localBuilder.Selector($"count(*)");
+                localBuilder.Selector("count(*)");
             }
 
             // Clear paging and order when counting 
--- a/src/YesSql.Core/Services/DefaultQuery.cs
+++ b/src/YesSql.Core/Services/DefaultQuery.cs
@@ -1046,7 +1046,7 @@ namespace YesSql.Services
             {
                 if (!_query._queryState._sqlBuilder.HasOrder)
                 {
-                    _query._queryState._sqlBuilder.OrderBy(_query._dialect.QuoteForColumnName("Id"));
+                    _query._queryState._sqlBuilder.OrderBy(_query._queryState._sqlBuilder.FormatColumn(_query._queryState._documentTable, "Id"));
                 }
 
                 _query._queryState._sqlBuilder.Skip(count.ToString());
@@ -1057,7 +1057,7 @@ namespace YesSql.Services
             {
                 if (!_query._queryState._sqlBuilder.HasOrder)
                 {
-                    _query._queryState._sqlBuilder.OrderBy(_query._dialect.QuoteForColumnName("Id"));
+                    _query._queryState._sqlBuilder.OrderBy(_query._queryState._sqlBuilder.FormatColumn(_query._queryState._documentTable, "Id"));
                 }
 
                 _query._queryState._sqlBuilder.Take(count.ToString());
--- a/src/YesSql.Core/Services/DefaultQuery.cs
+++ b/src/YesSql.Core/Services/DefaultQuery.cs
@@ -787,7 +787,7 @@ namespace YesSql.Services
 
             if (localBuilder.HasJoin)
             {
-                localBuilder.Selector($"count(distinct {_queryState._sqlBuilder.FormatColumn(_queryState._documentTable, "id")})");
+                localBuilder.Selector($"count(distinct {_queryState._sqlBuilder.FormatColumn(_queryState._documentTable, "Id")})");
             }
             else
             {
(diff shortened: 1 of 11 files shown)


wedge206:
I also fixed a typo on Attachments

wedge206:
Pushed a partial fix.  It fixes the two new unit tests, and keeps _almost_ everything else working.
Please sanity check these changes before I go much further down the rabbit hole.

src/YesSql.Provider.Common/SqlBuilder.cs
```diff
@@ -198,6 +207,17 @@ public virtual string ToSqlString()
 
                 sb.Append("SELECT ");
 
+                if (_distinct != null)
```
sebastienros: Looks like `_distinct` should be a bool and not a string

src/YesSql.Provider.Common/SqlBuilder.cs
```diff
@@ -198,6 +207,17 @@ public virtual string ToSqlString()
...
+                {
+                    sb.Append(_distinct);
+
+                    if (_order != null && _order[0] != "[Id]")
```
sebastienros: If this is to fix another unit test, I want to understand which one it is, because that seems super hacky

wedge206: It is hacky... and yes it's to fix another test.  I'm still working on it....

The problem is some queries do an Order By on Columns that aren't in the Select.  
But when DISTINCT used, then all Order By columns must be in the Select

src/YesSql.Provider.Common/SqlBuilder.cs
wedge206: Yeah, I couldn't decide on that.  I wasn't sure if some dialects used a different term or if Distinct is universal.  I'll make it a bool.

src/YesSql.Provider.Common/SqlBuilder.cs
wedge206: But I can't just add all the Order Clauses to the Select, because some break it in other ways

sebastienros: What database requires the select to have the order by field in the select clause?

wedge206: SQLServer does.  SQLite didn't seem to care.  I haven't been able to test other dbs locally

src/YesSql.Provider.Common/SqlBuilder.cs
```diff
@@ -198,6 +207,18 @@ public virtual string ToSqlString()
 
                 sb.Append("SELECT ");
 
+                if (_distinct)
+                {
+                    sb.Append("DISTINCT ");
+
+                    // This hacky bit needs a better solution
+                    if (_order != null && _order[0] != _dialect.QuoteForColumnName("Id"))
```
sebastienros: So actually the `_distinct` should be a string that contains the value as a parameter, optionally, like `""` to mean `DISTINCT` and `some thing` to mean `DISTINCT(some thing)`, and that something should be built using the dialect from outside the builder. Because here the `SqlBuilder` should not have to know about a column name (`[Id]`).

src/YesSql.Core/Services/DefaultQuery.cs
```diff
@@ -784,7 +784,15 @@ public async Task<int> CountAsync()
             var transaction = await _session.DemandAsync();
 
             var localBuilder = _queryState._sqlBuilder.Clone();
-            localBuilder.Selector("count(*)");
+
+            if (localBuilder.HasJoin)
+            {
+                localBuilder.Selector($"count(distinct {_queryState._sqlBuilder.FormatColumn(_queryState._documentTable, "id")})");
```
sebastienros: You are missing the table prefix.

src/YesSql.Provider.Common/SqlBuilder.cs
```diff
@@ -198,6 +207,18 @@ public virtual string ToSqlString()
...
+                    {
+                        _select.Add(",");
+                        _select.AddRange(_order.Where(o => o != " DESC"));
```
sebastienros: why filter by `DESC` ?

wedge206: Because we can't put 'DESC' into the SELECT clause.

wedge206: Turns out DISTINCT needs to know about a lot of things to work right.
One thought I had is there might need to be a new clause type.  So there would be "SELECT" and there would be "SELECT DISTINCT".  Then the dialects can have a little more control over it.

sebastienros: sorry, I thought the string contained both the name and "desc".
That will work, but still brittle as it expects the "strings" to be separate ...

sebastienros: or the builder could not have knowledge of distinct, and the caller to the builder would alter the select clause

src/YesSql.Provider.Common/SqlBuilder.cs
wedge206: I don't think I want to parameterize it yet.  It isn't needed to solve the current issue I'm trying to solve.  It would just create new problems to solve.

src/YesSql.Core/Services/DefaultQuery.cs
wedge206: I don't think it's missing, it seems to be working fine

src/YesSql.Core/Services/DefaultQuery.cs
```diff
@@ -1046,7 +1046,7 @@ IQuery<T> IQuery<T>.Skip(int count)
             {
                 if (!_query._queryState._sqlBuilder.HasOrder)
                 {
-                    _query._queryState._sqlBuilder.OrderBy(_query._dialect.QuoteForColumnName("Id"));
+                    _query._queryState._sqlBuilder.OrderBy(_query._queryState._sqlBuilder.FormatColumn(_query._queryState._documentTable, "Id"));
```
wedge206: Added the table name to these Order By Id lines, to resolve "Ambiguous column " errors when the order by gets added to the Select.
Honestly I'm surprised this worked before, if it's ambiguous in the Select, it should be ambiguous in the OrderBy too right?

But now it's causing an error in the Postgre test saying:
Column "tpDocument.Id" does not exist. But that doesn't make sense, that column definitely exists.
I'm currently unable to test Postgre locally, so any advice would be helpful

wedge206: Got it, Postgre is case sensitive.  I didn't know that.

sebastienros:
I am ok to merge it, unless you want to add something else

src/YesSql.Provider.SqlServer/SqlServerDialect.cs
```diff
@@ -157,9 +157,14 @@ public override void Concat(StringBuilder builder, params Action<StringBuilder>[
...
-            foreach(var o in orderBy.Where(o => o.Trim() != "," && o.Trim() != "DESC" && o.Trim() != "ASC"))
+            // Sql Server requires all Ordered fields to be part of the select when DISTINCT is used
+            
+            foreach(var o in orderBy)
             {
-                if (!select.Contains(o))
+                var trimmed = o.Trim();
+                
+                // Each order segment can be a field name, or a punctuation, so we filter out the punctuations 
+                if (trimmed != "," && trimmed != "DESC" && trimmed != "ASC" && !select.Contains(o))
```
hishamco: It would be better if it's like
```csharp
private static readonly string[] _orderList = { ",", "ASC", "DESC" }
...
if (!_orderList.Contains(trimmed) && !select.Contains(o))
```

sebastienros: I was also thinking of using a hashset in this case, but with 3 elements it's really equivalent to having a flat set of checks. So here it's really about code styling.

PR #209, opened 2019-10-20
Migrate Version Column

deanmarcussen:
Fixes an Orchard Core Issue https://github.com/OrchardCMS/OrchardCore/issues/4585 introduced with https://github.com/sebastienros/yessql/pull/208

The migration that was there doesn't run if the table exists.

This seemed like the best way to fix it. 
Run a single query on the document table and look for the version column
You may have better ideas, so change or do differently if you want :)

Only tested under sqlite, but I think we got the TOP / LIMIT issues on MS SQL fixed another time.

And it seems the tests pass second time around

### DIFF
--- a/src/YesSql.Core/Store.cs
+++ b/src/YesSql.Core/Store.cs
@@ -126,33 +126,34 @@ namespace YesSql
                         selectCommand.CommandText = selectBuilder.ToSqlString();
                         selectCommand.Transaction = transaction;
                         Configuration.Logger.LogTrace(selectCommand.CommandText);
-                        var result = await selectCommand.ExecuteReaderAsync();
-
-                        transaction.Commit();
-
-                        if (result != null && result.FieldCount == 4)
+                        using (var result = await selectCommand.ExecuteReaderAsync())
                         {
-                            return;
-                        }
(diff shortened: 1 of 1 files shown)


deanmarcussen:
On reflection this is not the best way to fix this for perf

`SELECT TOP 1 Version` (with the right dialects) returned as a scalar will throw at the right times, and the code that was there to create the table or throw, and add the version column.

I won't be able to amend until tonight

src/YesSql.Core/Store.cs
```diff
@@ -117,21 +117,53 @@ public async Task InitializeCollectionAsync(string collectionName)
...
+                                }
+                                return;
+                            }
                         }
                     }
                 }
-                catch
+                catch 
                 {
                     using (var transaction = connection.BeginTransaction())
```
sebastienros: Remove this inner catch

deanmarcussen: I think it needs it to handle the table create if the table itself doesn't exist?

sebastienros: right sorry, I thought it was the one I added, but you already removed it.

PR #223, opened 2020-01-11
Tests for index key length across providers

deanmarcussen:
@sebastienros you may/may not want to merge this, but it's a bunch of tests proving some of the MS Sql, and MySql index key length issues.

MySql limits to 784 total for strings under utf-8. Which decreases when bools area added, as it is both a individual and combined limit (underlying exception thrown is different tho)

MySql allows a prefix limit when creating the index.

MS Sql limits to 850, and again is a combined limit.
MS Sql doesn't offer the prefix limit, so in some ways that is the lowest common denominator we're stuck with (but without a builder no way to take advantage of the MySql prefix feature), so as below MySql then becomes the lowest common denominator

I did look at using a builder to use the MySql prefix limits, and MS Sql `include` statement, but dropped the idea, after further research indicated that the `include` statement meant the values where used only for `SELECT` but not for `WHERE` clauses, so would mean absolutely no index performance on `WHERE` for MS SQL on these larger index values.

Long and short of it is I think we're better of limiting some column values to a fixed length of 784 / 783 with booleans etc al, and truncating the other larger indexes (4000+) to within the MySql limits as they are the lowest.

So I'll adjust a couple of Orchard Core pr's to fit what these tests prove.


deanmarcussen:
ok, so further update.

I had to tweak appveyor to use mysql utf8 encoding to get the tests to fail.

utf8 encoding seems to be the recommended encoding for mysql, and mysql 8 will now use it by default (which is why tests failed locally).

With latin encoding (the old default) the value of a char is smaller, so the failing values are higher.

Everyone suggests that utf8 should be the encoding to use, but I am not a heavy mysql user, so I rely on advice of the internet.

Thoughts?

PR #232, opened 2020-05-20
Adding autoflush unit tests

sebastienros:
@jtkech


jtkech:
@sebastienros 

Okay, i was not able to update your branch directly, so i did the change that make the unit tests failing in #233, but where unit tests are not triggered.

PR #233, opened 2020-05-20
Make unit tests failing

jtkech:


jtkech:
Only doing `session.Query<Article, ArticlesByDay>(x => x.DayOfYear == 305)` doesn't do an auto flush.

PR #237, opened 2020-05-29
Use System.Text.Json APIs

### LINKED ISSUE #227: Any plan to support System.Text.Json?
Hi Sebastien ,

Do you have any plan to move to System.Text.Json for the serializer / deserializer?

Read about this and I think it could make YesSQL even faster.
https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-apis/

hishamco:
Fixes #227

### DIFF
--- a/src/YesSql.Core/Serialization/JsonContentSerializer.cs
+++ b/src/YesSql.Core/Serialization/JsonContentSerializer.cs
@@ -5,13 +5,20 @@ namespace YesSql.Serialization
 {
     public class JsonContentSerializer : IContentSerializer
     {
+        private readonly static JsonSerializerOptions _jsonSerializerOptions = new JsonSerializerOptions();
+
+        static JsonContentSerializer()
+        {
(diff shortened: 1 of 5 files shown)


hishamco:
This is WIP, because there 're some features not supported by JSON APIs, but I'm keeping to make the unit tests happy

hishamco:
FYI there're 6 failing tests due:

- Lack of support anonymous objects

- Lack of support non parameterless constructor

I'll struggle to make this tests pass

sebastienros:
And then you realize we can't use it in OC because we have custom Json.NET serialization attributes on content items.

It's still an interesting approach if the serializer is pluggable. Because some projects that don't rely on Json.NET. Or want to store object as binary (bson?, messagepack?)

Maybe some features could be disabled based on the serializer. The same way some sql databases don't support specific queries and disable a test altogether.

hishamco:
> It's still an interesting approach if the serializer is pluggable. Because some projects that don't rely on Json.NET. Or want to store object as binary (bson?, messagepack?)

It is cool if we made the serializer pluggable

PR #239, opened 2020-05-30
Introduce Naming Policy

### LINKED ISSUE #229: Change the names of the hardcoded Document table, index tables and the hard coded columns in config
Because I am using PostgreSQL, it is weird that the table names are in Pascal Case. This makes it harder to do queries directly to the database.
It would be better if there was an option to either configure the naming case, like PascalCase, camelCase, snake_case (which I would need here), or choose all the names ourselves.
I would first go for the first option, because this is the easiest to test (no need to test special cases of table names and protect for special characters, ...)

If I would add this fuctionality and PR it, would it be accepted? I don't really want to put time in something that won't be accepted anyways :)

hishamco:
Fixes #229


src/YesSql.Abstractions/PascalCaseNamingPolicy.cs
```diff
@@ -0,0 +1,22 @@
...
+        public override string ConvertName(string name)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentException("The name can't be null or empty.", nameof(name));
+            }
+
+            return string.Concat(name
+                .Split(_separator, StringSplitOptions.RemoveEmptyEntries)
+                .Select(w => w.Substring(0, 1).ToUpper() + w.Substring(1)));
```
hishamco, line 19: Need to be improved if it's possible

test/YesSql.Tests/NamingPolicyTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            var store = await StoreFactory.CreateAsync(configuration);
+            using (var connection = store.Configuration.ConnectionFactory.CreateConnection())
+            {
+                connection.Open();
+
+                using (var transaction = connection.BeginTransaction(store.Configuration.IsolationLevel))
+                {
+                    var builder = new SchemaBuilder(store.Configuration, transaction);
+                    builder.CreateTable("UsersInRoles", column => column
+                            .Column<int>(store.Configuration.NamingPolicy.ConvertName("UserId"))
```
hishamco, line 31: Need to find a way to apply naming policy nicely with column name(s)

src/YesSql.Core/Services/DbBlockIdGenerator.cs
```diff
@@ -43,9 +43,9 @@ public async Task InitializeAsync(IStore store, ISchemaBuilder builder)
...
-            UpdateCommand = "UPDATE " + _dialect.QuoteForTableName(_tablePrefix + TableName) + " SET " + _dialect.QuoteForColumnName("nextval") + "=@new WHERE " + _dialect.QuoteForColumnName("nextval") + " = @previous AND " + _dialect.QuoteForColumnName("dimension") + " = @dimension;";
-            InsertCommand = "INSERT INTO " + _dialect.QuoteForTableName(_tablePrefix + TableName) + " (" + _dialect.QuoteForColumnName("dimension") + ", " + _dialect.QuoteForColumnName("nextval") + ") VALUES(@dimension, @nextval);";
+            SelectCommand = "SELECT " + _dialect.QuoteForColumnName("nextval") + " FROM " + _dialect.QuoteForTableName(_tablePrefix + store.Configuration.NamingPolicy.ConvertName(TableName)) + " WHERE " + _dialect.QuoteForColumnName("dimension") + " = @dimension;";
+            UpdateCommand = "UPDATE " + _dialect.QuoteForTableName(_tablePrefix + store.Configuration.NamingPolicy.ConvertName(TableName)) + " SET " + _dialect.QuoteForColumnName("nextval") + "=@new WHERE " + _dialect.QuoteForColumnName("nextval") + " = @previous AND " + _dialect.QuoteForColumnName("dimension") + " = @dimension;";
```
sebastienros, line 47: If the naming policy was part of the dialect, then we wouldn't have to change most code as it would be automatic when QuoteForXYZ is called.

But that would mean to have an overload with QuoteForTableName(string prefix, string name) so we can only convert the table name and not the prefix.

test/YesSql.Tests/NamingPolicyTests.cs
sebastienros: User code should not have to deal with it, it should be part of the schema builder logic, the same way it will add the table prefix by itself.

src/YesSql.Core/Configuration.cs
```diff
@@ -23,18 +23,31 @@ public Configuration()
             QueryGatingEnabled = true;
             Logger = NullLogger.Instance;
             ConcurrentTypes = new HashSet<Type>();
+            NamingPolicy = NamingPolicy.PascalCase;
```
sebastienros, line 26: The default naming policy should be a NullNamingPolicy one. And keep whatever case is set by the user.

hishamco: > The default naming policy should be a NullNamingPolicy one

Ya, but if you look closely we're using PacalCase by default

src/YesSql.Core/Services/DbBlockIdGenerator.cs
hishamco: I was thinking about that before, but these methods are `abstract`, so let me revise the code and see what should I come up with

src/YesSql.Core/Configuration.cs
hishamco: > User code should not have to deal with it, it should be part of the schema builder logic, the same way it will add the table prefix by itself.

I already did that for non columns, but I stuck in `CreateTable` because it passes the columns in the action delegete

src/YesSql.Core/Services/DbBlockIdGenerator.cs
hishamco: > But that would mean to have an overload with QuoteForTableName(string prefix, string name) so we can only convert the table name and not the prefix.

I realize that there 're some places will not work, on example `CreateTableCommand` because it pass a table name with a prefix to the dialect

hishamco:
@sebastienros your final review, one more thing I'd like to mention, I prefer to use `PascalCase` instead of `Null` coz that what is the previous case for table names

sebastienros:
It's about perf. Null would not allocate anything in the default case.

hishamco:
@sebastienros could you merge this?

hishamco:
Seb there's no plan to support naming policy?

hishamco:
Seb there's no plan to support naming policy?

hishamco:
Seb there's no plan to support naming policy?

sebastienros:
I renamed the default branch to main and it closed all PRs. But even though I don't think I want to take the increased complexity.

Issue #240, opened 2020-06-01
Error commiting changes in two collections


jersiovic:
Hello,

I've managed to separate in a diffent Document tokens and authorizations of opeind module. The purpose is to solve the issue mentioned here: https://github.com/OrchardCMS/OrchardCore/issues/6318

For doing that I provide my own stores for those two entities. Also I use this simple wrapper of ISession to avoid having to use have all the code filled with usings for specify everywhere the collection for the operation to yessql. 

```
public class CollectionSession : ISession
    {
        public CollectionSession(string collectionName, ISession session)
        {
            Session = session;
            CollectionName = collectionName;
        }
        public string CollectionName { get; }
        public ISession Session { get; }
        
        public IStore Store => Session.Store;

        public void Cancel()
        {
            using (new NamedCollection(CollectionName))
            {
                Session.Cancel();
            }
        }

        public async Task CommitAsync()
        {
            using (new NamedCollection(CollectionName))
            {
                await Session.CommitAsync();
            }
        }

        public void Delete(object item)
        {
            using (new NamedCollection(CollectionName))
            {
                Session.Delete(item);
            }
        }
        ....
        and so on
```
So the only thing that changes at the original OpenIdTokenStore is this:
```
public OpenIdTokenStore(ISession session)
        {
            _session = new CollectionSession(ModuleCollectionNames.OpenIdCollectionName, session);
        }
```
I ve also override the default Orchard registration of IStore to be able of registering indexes for another collection different from defaul Orchard one. In this way:
```
...
                var indexesPerCollection = sp.GetServices<IIndexProvider>().GroupBy(i => i.CollectionName);
                foreach (var indexes in indexesPerCollection)
                {
                    if (string.IsNullOrWhiteSpace(indexes.Key))
                    {
                        store.RegisterIndexes(indexes);
                    }
                    else
                    {
                        using (new NamedCollection(indexes.Key))
                        {
                            store.RegisterIndexes(indexes);
                        }
                    }
                }
...
```
This is possible cause YesSql IIndexProvider provides a property CollectionName that I initialize with the name of the collection for those indexes that have to work with a different collection.

The point with that is in this way we can use YesSql collections easily and with low impact on any  Orchard module.

Problem is the way Orchard manages ISession and how YesSql expect we store data in collections in separate commits. Because it is difficult to be sure that one commit in Orchard only will change data from one collection.

[...]

PR #241, opened 2020-06-01
Allow commit on different collections

jersiovic:
PR related with this issue https://github.com/sebastienros/yessql/issues/240

### DIFF
--- a/src/YesSql.Core/Session.cs
+++ b/src/YesSql.Core/Session.cs
@@ -23,20 +23,20 @@ namespace YesSql
         internal readonly List<IIndexCommand> _commands = new List<IIndexCommand>();
         private readonly Dictionary<IndexDescriptor, List<MapState>> _maps = new Dictionary<IndexDescriptor, List<MapState>>();
         
-        // entities that need to be created in the next flush
-        private readonly HashSet<object> _saved = new HashSet<object>();
+        // entities that need to be created in the next flush on a collection
+        private readonly Dictionary<object, string> _saved = new Dictionary<object, string>();
 
-        // entities that already exist and need to be updated in the next flush
-        private readonly HashSet<object> _updated = new HashSet<object>();
+        // entities that already exist and need to be updated in the next flush on a collection
+        private readonly Dictionary<object, string> _updated = new Dictionary<object, string>();
 
-        // entities that are already saved or updated in a previous flush
-        private readonly HashSet<object> _tracked = new HashSet<object>();
+        // entities that are already saved or updated in a previous flush on a collection
+        private readonly Dictionary<object, string> _tracked = new Dictionary<object, string>();
 
         // ids of entities that are checked for concurrency
         private readonly HashSet<int> _concurrent = new HashSet<int>();
 
         // entities that need to be deleted in the next flush
-        private readonly HashSet<object> _deleted = new HashSet<object>();
+        private readonly Dictionary<object, string> _deleted = new Dictionary<object, string>();
         protected readonly Dictionary<string, IEnumerable<IndexDescriptor>> _descriptors = new Dictionary<string, IEnumerable<IndexDescriptor>>();
         internal readonly Store _store;
         private volatile bool _disposed;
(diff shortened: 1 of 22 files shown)


src/YesSql.Core/Session.cs
```diff
@@ -780,17 +784,19 @@ private void BatchCommands()
 
             if (createDocumentCommands != null)
             {
-                foreach (var page in createDocumentCommands.PagesOf(_store.Configuration.CommandsPageSize))
+                foreach (var page in createDocumentCommands.PagesOf(_store.Configuration.CommandsPageSize)
+                    .SplitPagesBy(c=>c.CollectionName))
```
jersiovic, line 788: This call to SplitPagesBy is one thing I don't feel pretty well.
The poins is I saw that you pass to dapper an string with an insert in the pseudo SQL and a collection of documents to insert. Problem I found is the SQL is not the same for all the documents because some of them are for one collection and others are for others one.
I had two options: 

- To group inserts (or updates on next foreach) to make as much as possible inserts in one batch commad. But if generates me doubts cause I'm not sure it is right to change the oreder on which things are inserted on db.
- To split each page every time I find a document with a different collection than previous one. Problem is this breaks  batch in small pieces.

What do you think, is it ok like it is?
Thank you

src/YesSql.Core/Session.cs
```diff
@@ -780,17 +784,19 @@ private void BatchCommands()
...
-                    _commands.Add(new CreateDocumentCommand(page.SelectMany(x => x.Documents), _tablePrefix));
+                    _commands.Add(new CreateDocumentCommand(page.First().CollectionName, page.SelectMany(x => x.Documents), _tablePrefix));
                 }
             }
 
             if (deleteDocumentCommands != null)
             {
-                foreach (var page in deleteDocumentCommands.PagesOf(_store.Configuration.CommandsPageSize))
+                foreach (var page in deleteDocumentCommands.PagesOf(_store.Configuration.CommandsPageSize)
+                    .SplitPagesBy(c => c.CollectionName))
```
jersiovic, line 797: same here

sebastienros:
Notes from meeting:

We need index tables for each document table (each collection) because of foreign keys

Query<Token, TokenIndex>(collection)
Query<TokenIndex>(collection)

Document
TokenIndex

Collection1_Document
Collection1_TokenIndex

store.RegisterIndexProvider<P>(collection)

{ _saved, _updated, _delete } should be a set per collection
indexProviders should be a list per collection

jersiovic:
I've updated this PR with the changes I made to make that any table created, updated or queried uses as prefix, the prefix and the collection name.
So no matter if you are creating a regular table, a document table or an index table that prefix will be added.
This PR now has what I'm going to use in production together with the mechanism I was using at Orchard side.
I provide it to you only just in case it could help you to spot points that will need to be changed when you start working on what we agreed on the meeting.

Appveyor show 3 errors for postgress tests .. I will install Postgress to check what happens

jersiovic:
Fixed on last commit also for Postgress

jersiovic:
Added another commit on this PR, cause as it was it didn't work with my approach of having my own Session and SchemaBuider enveloping original objects when the type or the type name are related to a different collection prefix.
Problem was that queries like this one didn't work (without using an `using Collection collection= ...) enveloping it`:
`var users = await _session.Query<User, UserByClaimIndex>(uc => uc.ClaimType == claim.Type && uc.ClaimValue == claim.Value).ListAsync();`

My Session object (called CollectionSession)I retuns a Query enveloped  in my own ColletionQuery object.
```
...
        public IQuery Query()
        {
            return new CollectionQuery(Session.Query(), _yesSQLCollectionNameService);
        }
...
```
whic also envelops ForEach calls within a collection
```
public IQuery<T> For<T>(bool filterType = true) where T : class
        {
            var collectionName = _yesSQLCollectionNameService.GetCollectionNameForType(typeof(T));
            if (collectionName == null)
            {
                return _query.For<T>(filterType);
            }
            using (new NamedCollection(collectionName))
            {
                return _query.For<T>(filterType);
            }
        }
```
However this is not enough cause when at YesSql DefaultQuery evaluates current collection when the query is created, it still doesn't know which is the type rquested in the query. So I need to delay this check to the momento For method is called in DefaultQuery.
What I did is hacky and has potential problems, that by the moment I've seen are not happening to me. So I'm going to use it temporally while you work on the final solution that allows relate entities with collections within YesSQL

PR #246, opened 2020-06-22
Collections refactoring

### LINKED ISSUE #240: Error commiting changes in two collections

sebastienros:
- Exposing collections in the main APIs
- Support for multi-collections transactions

Fixes #240


sebastienros:
@jersiovic Are you able to test this branch? All tests are passing.
I want to add one more unit test for what I think is a bug I discovered while working on this, but it should already work for your needs.

PR #261, opened 2020-08-10
Add default precision and scale for dialects and respect the passed one for decimal values.

asimeonov:
Currently yessql don't respect passed values for precision and scale for decimals. 

This PR adds default values per dialect and changes precision and scale properties to nullable (maybe a braking change?)


sebastienros:
Looks good. Could you add a unit test also to show what the intended usage is, and also verify it works on all db providers with common values.

asimeonov:
Yeah will do it in the begging of next week @sebastienros when I go back from vacation.

asimeonov:
@sebastienros I have create simple test that verifies the output of `GetTypeName` method. If this is not sophisticated enough I can use something like 
`SELECT NUMERIC_PRECISION as [PRECISION], NUMERIC_SCALE as [SCALE]
FROM INFORMATION_SCHEMA.COLUMNS
WHERE 
     TABLE_NAME = 'SomeTableWithDecimalColumn' AND 
     COLUMN_NAME = 'DecimalColumnName'`  

for SQL Server and respective implementation for the rest of Dialects.
Also the build is failing for some reason that I didn't quire understand. I'm not sure is it related to my changes considering that all tests are passing.

PR #275, opened 2020-10-02
Adding support for IN filters without predicate

### LINKED ISSUE #274: Consider adding an IsNotIn overload that doesn't require a second argument
Such an overload would be useful in the OrchardCore OpenID module: https://github.com/OrchardCMS/OrchardCore/pull/7193#discussion_r498947921

sebastienros:
Fixes #274

You can now either:
- use `IsInAny` 
- pass `null` as the predicate
- pass `x => true`

/cc @kevinchalet


kevinchalet:
Thanks! 🎉

PR #277, opened 2020-10-09
Adding synchronous methods to create a store

### LINKED ISSUE #188: Make Store constructor public
I would like to asynchronously initialize the `Store` instead of doing it using a blocking call from the service registration step (using [asynchronous initialization](https://thomaslevesque.com/2018/09/25/asynchronous-initialization-in-asp-net-core-revisited/)).

Unfortunately, the only way to to construct a `Store` is through `StoreFactory`.

What I'd like to do is:

1. Register `Store` as a singleton (uninitialized)
2. Use async initialization to initialize the store (which will happen before the app starts, so it's all good)

If you agree I'll send a PR.

sebastienros:
Fixes #188

### DIFF
--- a/samples/YesSql.Bench/Program.cs
+++ b/samples/YesSql.Bench/Program.cs
@@ -37,7 +37,7 @@ namespace Bench
                 }
             }
 
-            var store = await StoreFactory.CreateAsync(configuration);
+            var store = await StoreFactory.CreateAndInitializeAsync(configuration);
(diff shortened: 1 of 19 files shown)


hishamco:
Seb this introduces a breaking changes, so can bump the version?

sebastienros:
I know, but no ;)

hishamco:
As you wish ;)

PR #278, opened 2020-10-10
Adding AlterIndexTable

### LINKED ISSUE #264: Collections with AlterTable command using nameof
I haven't looked too far into this but trying to establish whether the collections / table altering is solid.

We can create a MapIndex table like so
`SchemaBuilder.CreateMapIndexTable<AutoMenuPartIndex>(table => table`

but when going to add an index to it the table name still takes a string
`SchemaBuilder.AlterTable(nameof(AutoMenuPartIndex), table => table`

That's ok, because I can assume that `CreateMapIndexTable<AutoMenuPartIndex>` will create a table named `AutoMenuPartIndex` if not using a collection, but I'm assuming this to be the case.

Seems strange that I'm forced on a type to one, which might be impacted by a naming policy, but not on the other.

Equally it's still probably useful to alter a table based on it's name

sebastienros:
Fixes #264

### DIFF
--- a/samples/YesSql.Samples.Web/Startup.cs
+++ b/samples/YesSql.Samples.Web/Startup.cs
@@ -12,17 +12,18 @@ namespace YesSql.Samples.Web
             services.AddDbProvider(config =>
                 config.UseSqlServer("Server=.;Database=YesSqlDb;Integrated Security=True"));
 
-            services.AddMvc();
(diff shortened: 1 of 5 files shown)


sebastienros:
@deanmarcussen looks good?

deanmarcussen (approved):
Merge it ;)

PR #280, opened 2020-10-10
Fix that Indexes and ScopedIndexes are null upon Store construction

sfmskywalker:
When instantiating a new `Store`, it is impossible to add index providers to `Indexes` until after `InitializeAsync` is called.
This PR fixes that by making sure `Indexes` (and `ScopedIndexes`) is initialized to an empty list as soon as the `Store` is instantiated (rather than waiting to do so during `InitializeAsync`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ project.lock.json
 .vs/
 .build/
 .testPublish/ 
+.idea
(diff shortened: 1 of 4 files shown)


sfmskywalker:
I also changed a few other files. Very small changes, but I can create a new PR with just the `Store` change if you want.

PR #287, opened 2020-11-07
Check concurrency issue

jtkech:
Related to https://github.com/OrchardCMS/OrchardCore/issues/7449 and https://github.com/OrchardCMS/OrchardCore/issues/7545

**Repro**

- Save a given entity with `checkConcurrency` to true
- Do a first session query non related to the above entity => `FlushAsync()`
- Do at least a 2nd session query => `FlushAsync()`
- ...
- Save again the same entity with `checkConcurrency` to true
- `CommitAsync()`

On the first `FlushAsync()` a `SaveEntityAsync()` is done that enlists a `CreateDocumentCommand()` that is then executed, finally the entity is added to `state.Tracked` and removed from `state.Saved`. On the second `FlushAsync()`, because the entity is in `state.Tracked`, `UpdateEntityAsync()` is called where, because of the `checkConcurrency`, **the version is incremented but, because the entity is tracked and its content has not changed yet, the related `UpdateDocumentCommand` is not enlisted, so not executed through the current flush**.

Then when the entity is saved again, it is removed from `state.Tracked` and added to `state.Updated`, so when the session is committed, through the final flush that recalls `UpdateEntityAsync()`, **the version is incremented again** and, because the entity is no more tracked (and its content may have been updated), an `UpdateDocumentCommand` is enlisted and then executed. **So the document version is incremented twice but only one `UpdateDocumentCommand` is executed, so there is a version missmatch resulting in a false `ConcurrencyException`**

So here,  if the entity is tracked and its content has not changed yet, i revert the document version update.

**Update**:

I built YesSql locally and made packages in a local feed, so i could try my YesSql updates, I can confirm that it fixes the related OC issue. Note: To be able to build OC in this context i needed to do some adaptations as currently we don't reference the last dev of YesSql. E.g. `StoreFactory.CreateAsync()` replaced by `StoreFactory.CreateAndInitializeAsync()`.

**Update**: I found a temporary fix on OC side, defining a `Version` property in our `Document` base class prevents a version missmatch, this because the version of the tracked object is also incremented, so its serialized content becomes different and then the related `UpdateDocumentCommand` is enlisted.

### DIFF
--- a/src/YesSql.Core/Session.cs
+++ b/src/YesSql.Core/Session.cs
@@ -370,7 +370,7 @@ namespace YesSql
                 {
                     oldDoc.Version--;
 
-                    // restore the old version to the object
+                    // revert the version update
                     if (versionAccessor != null)
                     {
                         versionAccessor.Set(entity, (int)oldDoc.Version);
--- a/src/YesSql.Core/Session.cs
+++ b/src/YesSql.Core/Session.cs
@@ -124,7 +124,7 @@ namespace YesSql
                 {
                     state.IdentityMap.AddEntity(id, entity);
                     state.Updated.Add(entity);
-                    
+
                     // If this entity needs to be checked for concurrency, track its version
                     if (checkConcurrency || _store.Configuration.ConcurrentTypes.Contains(entity.GetType()))
                     {
@@ -223,7 +223,7 @@ namespace YesSql
         public void Detach(object entity, string collection)
         {
             CheckDisposed();
-            
+
             var state = GetState(collection);
 
             state.Saved.Remove(entity);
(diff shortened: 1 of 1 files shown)


sebastienros:
Thanks! I might steal this PR and do a unit test first, to be sure we don't regress, and understand how it's happening to be sure there is no better solution than call `--` which scares me. Like is there any way that the version could be incremented more than once?

jtkech:
> Like is there any way that the version could be incremented more than once?

Yes, on a first flush for a saved object that is then marked as tracked, then in a 2nd flush because it is tracked, but if its content was not changed the related update command is not enlisted. So in the last commit there is a version mismatch.

As said i found a fix on OC side by adding a `Version` property to our `Document` base class. So in this case this property is also incremented, so we see that the document content has changed and then the related update command is enlisted.

---

But doing this results in more sql cmds for tracked objects (if multiple flushs), that will be fixed with this PR and removing the `Version` property on OC side, but i also thought about objects for which people still want to have a version property

So, and only for a tracked object, maybe comparing the `newContent` with `oldDoc.Content` before incrementing the version, but i didn't go further because if the content are not equal, we would need a 2nd serialization after incrementing the version.

PR #305, opened 2020-12-02
SqlServer 2008 Dialect fixing pagination with nested query

jersiovic:
Just in case it is useful to merge or to combine with this other PR https://github.com/sebastienros/yessql/pull/64 . Here is my proposal for fixing Sql Server 2008 pagination. 
It produces nested queries for this purpose cause as far as I know this is the only way of having pagination with this sql server version.
This is the kind of query it is produces:

SELECT * FROM (SELECT DISTINCT [Document].*, ROW_NUMBER() OVER( ORDER BY [UserIndex].[NormalizedUserName]) AS Seq,[UserIndex].[NormalizedUserName] FROM [Document] INNER JOIN [UserIndex] ON [UserIndex].[DocumentId] = [Document].[Id] ) yessqlQuery WHERE Seq BETWEEN 1 AND 10

I've checked pagination tests run ok with this dialect and Sql Server 2008
However other tests don't pass.

The failing tests are a group related with index lenght as @deanmarcussen pointed here https://github.com/sebastienros/yessql/pull/64#issuecomment-734159110

````
ShouldCreateAndIndexPropertyWithMaximumKeyLengths Failed	7,4 sec		Microsoft.Data.SqlClient.SqlException : Operation failed. The index entry of length 1535 bytes for the index 'IDX_Property' exceeds the maximum length of 900 bytes.
````
````
ShouldIndexPropertyKey Failed	7,2 sec		Microsoft.Data.SqlClient.SqlException : Operation failed. The index entry of length 1700 bytes for the index 'IDX_Property' exceeds the maximum length of 900 bytes.
````
````
ShouldIndexPropertyKeys Failed	7 sec		Microsoft.Data.SqlClient.SqlException : Operation failed. The index entry of length 1700 bytes for the index 'IDX_Property' exceeds the maximum length of 900 bytes.
````
````
ShouldIndexPropertyKeysWithBits Failed	9,6 sec		Microsoft.Data.SqlClient.SqlException : Operation failed. The index entry of length 1699 bytes for the index 'IDX_Property' exceeds the maximum length of 900 bytes.
````

IMO, this kind of errors is not a problem that yessql should control, it is a problem for those who use YesSql with a version of a db with a smaller maximum leght for the index. So, they will have to limit the lenght of the indexes defined at their code if they want to support versions with a smaller maximum leght. For example maybe Orchard Core should be sure to don't use indexes greater that the minimum of SqlServer 2008 on its code. Developers will be free of using the size of the index they want only on their custom modules.

Another group of test errors happens on tests that are sensible to the delay in communication with db:
```
 ShouldGenerateIdsConcurrently(collection: "")
   Source: CoreTests.cs line 3984
   Duration: 17,6 sec  Message: 
    lastId: 745
    Expected: True
    Actual:   False
```
```
 ShouldGenerateIdsConcurrently(collection: "Collection1")
   Source: CoreTests.cs line 3984
   Duration: 17,6 sec

  Message: 
    lastId: 825
    Expected: True
    Actual:   False
```
Reason is I'm using an sql server 2008 hosted in a cheap virtual machine at the cloud for testing from my local pc, so I wouldn't worry of those fails.

Finally the last group of errors are those caused by an error that exist in dev branch which makes failing these tests also for SQL Server 2019 with original SQLServer dialect, because the wrong connection string is used for them. As soon as they are fixed for original dialect I expect it will work for SQLServer2008 dialec:
```
 ShouldGenerateIdsWithConcurrentStores(collection: "")
   Source: SqlServerTests.cs line 97
   Duration: 36 sec

  Message: 
    Microsoft.Data.SqlClient.SqlException : A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server)
````
````
 ShouldGenerateIdsWithConcurrentStores(collection: "Collection1")
   Source: SqlServerTests.cs line 97
   Duration: 34,9 sec

  Message: 
    Microsoft.Data.SqlClient.SqlException : A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server)
````

### DIFF
--- a/src/YesSql.Provider.Common/SqlBuilder.cs
+++ b/src/YesSql.Provider.Common/SqlBuilder.cs
@@ -24,6 +24,8 @@ namespace YesSql.Sql
         protected bool _distinct;
         protected string _skip;
         protected string _count;
+        protected List<string> _parentSelect;
+        protected List<string> _parentWhere;
 
 
         protected List<string> SelectSegments => _select = _select ?? new List<string>();
@@ -35,6 +37,8 @@ namespace YesSql.Sql
         protected List<string> HavingSegments => _having = _having ?? new List<string>();
         protected List<string> OrderSegments => _order = _order ?? new List<string>();
         protected List<string> TrailSegments => _trail = _trail ?? new List<string>();
+        protected List<string> ParentQuerySelectSegments => _parentSelect = _parentSelect ?? new List<string>();
+        protected List<string> ParentQueryWhereSegments => _parentWhere = _parentWhere ?? new List<string>();        
 
         public Dictionary<string, object> Parameters { get; protected set; } = new Dictionary<string, object>();
 
@@ -144,6 +148,39 @@ namespace YesSql.Sql
             }
         }
 
+        public void ParentSelector(string selector)
+        {
+            ParentQuerySelectSegments.Clear();
+            ParentQuerySelectSegments.Add(selector);
+        }
+
+        public void ParentSelector(string table, string column)
+        {
+            ParentSelector(FormatColumn(table, column));
+        }
+
+        public void AddParentSelector(string select)
+        {
+            ParentQuerySelectSegments.Add(select);
+        }
+
+        public void InsertParentSelector(string select)
+        {
+            ParentQuerySelectSegments.Insert(0, select);
+        }
+
+        public string GetParentSelector()
+        {
+            if (ParentQuerySelectSegments.Count == 1)
+            {
+                return ParentQuerySelectSegments[0];
+            }
+            else
+            {
+                return string.Join("", ParentQuerySelectSegments);
+            }
+        }
+
         public void Distinct()
         {
             _distinct = true;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/YesSql.Provider.Common/SqlBuilder.cs [hunk 1]
```diff
@@ -349,7 +433,7 @@ public virtual string ToSqlString()
                     }
                 }
 
-                if (_order != null)
+                if (_order != null && _parentSelect == null)
```
jersiovic, line 436: An alternative to avoid this difficutl to understand ´&& _parentSelect == null` we could add a method to SqlBuilder which allow us to set to null the `_order` property from the SQLServer2008 dialect class when pagination is requested.
What do you think?

deanmarcussen:
> IMO, this kind of errors is not a problem that yessql should control

Yes, I kind of agree @jersiovic 

For infos

I wrote those tests to provide a baseline, mostly for OrchardCore, but on the basis that YesSql supports swappable database providers, so it should have a baseline of support.

The key length limits are from 2016 +, so 2014 and below would be the old key size limit, as you have identified.

MySql also ended up providing the lowest limit, but it's limit is still higher than SQL 2014...

All good from my point of view to rearrange the tests for a lower baseline, but we just wouldn't be able to bring the SQL2008 provider into OrchardCore, without having to rearrange indexes and/or max lengths of some fields.

sebastienros:
I'd rather pay for your customer to upgrade to a better sql version that have to maintain this code and all the complexities it will add everywhere.

jersiovic:
> I'd rather pay for your customer to upgrade to a better sql version than have to maintain this code and all the complexities it will add everywhere.

That was just my first answer when we noticed the issue. Problem is we found this is sth that could affect an undefined number of our customers, That's why I made the effort of providing this PR despite it increases the complexity.

jersiovic:
Another option that wil work for me is discard SqlServerDbProviderOptionsExtensions.cs  and SqlServer2008Dialect.cs files of this PR, but accept the other changes or other similar to those. In this way you don't support officially SQL Server 2008 but provide extension points for those willing to provide its own SqlServerDbProviderOptionsExtensions and dialect for supporting it out of what YesSQL support

jersiovic:
I close this PR as it doesn't seems it will be merged

Issue #311, opened 2020-12-16
Column with the Guid type is created as a string


MikeAlhayek:
I tried to create a database table using a migration with the following method.

```
SchemaBuilder.CreateTable("Vendors", (table) => table
    .Column<Guid>("Id", column => column.PrimaryKey().WithLength(36))
    .Column<string>("Name", (column) => column.WithLength(100).NotNull())
);
```
The above command creates a column of type `Guid` as nvarchar(36) instead of [UniqueIdentifier](https://github.com/sebastienros/yessql/blob/dev/src/YesSql.Provider.SqlServer/SqlServerDialect.cs#L13)

I am able to workaround this problem like this, but think there

```
SchemaBuilder.CreateTable("Vendors", (table) => table
    .Column("Id", DbType.Guid, column => column.PrimaryKey())
    .Column<string>("Name", (column) => column.WithLength(100).NotNull())
);
```

If this is agreed to be a problem, I think the problem is in [this line](https://github.com/sebastienros/yessql/blob/68972b2325c0594f98491fd40e2c77bd75dd3b59/src/YesSql.Core/Sql/Schema/SchemaUtils.cs#L28) which should be changed to `{ typeof(Guid), DbType.Guid}`

sebastienros:
This should be fixed in the latest version. Can you check? There are unit tests too. It's using a string only for the databases that don't support it.

CrestApps:
@sebastienros Yes that indeed fixed the issue. 

I noticed you changed `DbType` to `Type` and you removed the `ISqlDialect Dialect` from the `IStore`. To get the name of the server being used ("MySQL","SqlServer"....) is it valid to do this now?

```
IStore store = serviceProvider.GetRequiredService<IStore>();
string sqlServerName = store.Configuration.SqlDialect.Name;
```

PR #312, opened 2020-12-16
Update SchemaUtils.cs

CrestApps:
Guid should return Guid not string. Fixing #311


sebastienros:
Would you be able to add a unit test that uses a field of this type? Like add a Guid property to an existing index, and assign one to it and check that it can be used in a query to filter?

CrestApps:
@sebastienros are you looking for a [test like this](https://github.com/CrestApps/yessql/blob/dev/test/YesSql.Tests/SqlServerTests.cs#L446-L535) and if so, so you want me to create another pull request?

test/YesSql.Tests/SqlServerTests.cs
```diff
@@ -442,5 +442,96 @@ public async Task ShouldIndexPropertyKeysWithBits()
                 session.Save(property);
             }
         }
+
+        [Fact]
+        public async Task AbleToCreateGuidAsUniqueIdentifier()
```
sebastienros, line 447: What if the test is in CoreTests instead? Such that it's tested on all database and not just SQL Server?
I assume that's why it was mapped to a string in the first place. Or the solution should take that into account and behave different on each database, maybe by having type mappings defined in the dialect.

CrestApps: Yes I can move it up. I just don't have instance of every database to run the test on. But, I can move it to CoreTests. Would you want me to send it a second pull request and you can run a full test on your side?

sebastienros: Just move it up in this PR, the CI will do it for you. I don't have any database on my local system ;) I rely on the PR CI.

CrestApps: @sebastienros as you guessed the test failed on one of the database providers :( I was able to run the test on MySQL and SQL Server database and it worked. I am guessing it is not working on other providers. Not sure if there is a way to tell which provider it is failing for. Is there a way to fix this issue so that Guid is treated as Guid not as a string?

sebastienros: It's in the build log on the CI, where you took the message from. In this case Sqlite. If it had passed it would have then tried with PostgresQL. 
It might require to move the dictionary you changed to the Dialect base class, and let each dialect define custom mappings for the types. That was expected.
You can try it if you want, it shouldn't be complex since I assume all the code paths that need the dictionary already have access to the dialect.

sebastienros:
Now Dapper might require a custom type handler

sebastienros:
https://stackoverflow.com/a/52319934/142772

sebastienros:
Unfortunately, this is a static collection, which means this won't work when using multiple databases in the same app, or in a multitenant scenario in Orchard Core. Because the other connections don't want that conversion

sebastienros:
It might be possible to convert a Guid for each db provider by adding an ITypeHandler implementation: https://github.com/StackExchange/Dapper/blob/6ab170c160f0689e2e2bf7e6eb603447f734d454/Dapper/UdtTypeHandler.cs#L12

The methods could maybe check the type of the parameter, and then convert correctly. It would use another static dictionary that each provider would register into to provide their own handlers. And only Sqlite would provide a custom one.

CrestApps:
@sebastienros It looks like my last changes fixed the issue with Sqlite. I add [RegisterTypeHandlers](https://github.com/CrestApps/yessql/blob/dev/src/YesSql.Provider.Common/BaseDialect.cs#L35-L38) then called it from the [CreateBuilder](https://github.com/CrestApps/yessql/blob/dev/src/YesSql.Provider.Common/BaseDialect.cs#L222) and finally add this to [SqliteDialect](https://github.com/CrestApps/yessql/blob/dev/src/YesSql.Provider.Sqlite/SqliteDialect.cs#L51-L56)

I don't know if this is a valid approach or not. I am not sure if calling the RegisterTypeHandlers() from in the `CreateBuilder` is a good idea but not sure what other options I can try.

Assuming `RegisterTypeHandlers()` is an acceptable approach, I [PostgreSql provider now throwing this error](https://ci.appveyor.com/project/SebastienRos/yessql-un1yf/builds/36876385#L185)
 > Npgsql.PostgresException : 42883: operator does not exist: character = uuid

CrestApps:
@sebastienros I changed the approach I mentioned above. Instead, I created a method called [Make](https://github.com/CrestApps/yessql/blob/dev/src/YesSql.Provider.Sqlite/SqliteDialect.cs#L51-L58) that would allow us to instantiate the class and register the type. Not sure of a better way.

Either way the CI is still failing Postgre test is failing this check [ShouldReadUncommittedRecords](https://github.com/CrestApps/yessql/blob/dev/test/YesSql.Tests/CoreTests.cs#L3185). Can you help me with this error?

PR #338, opened 2021-01-21
Removing usage of DynamicParameters

sebastienros:


deanmarcussen:
Ok, so in regard to https://github.com/sebastienros/yessql/issues/339

It's probably better to do the changes on this pr.

I looked and I don't think `CommandsPageSize` is appropriate, as one can't know how many parameters are in a command before hand when preparing the page. (maybe ?)

But looks like it would be pretty simple to put something on the dialect / `AddParameters` extension which would return false, and start a new batch, if the number of parameters hit 2100

PR #340, opened 2021-01-24
Reduce max commands to 2098

deanmarcussen:
This first run will fail.

This query, propitiously, produces a batch of 2099 exactly, on its first batch.

And sql server will drop it, even though it doesn't exceed 2100. 
I wonder if it is adding an extra param or two, from the `SCOPE_IDENTITY` but not sure.

It's not a perfect test, because if the model or index changed, and it wasn't producing a batch of 2099, it would still pass,  (some experimentation with other models in here as well changed the batched parameter count to 2097, or 2096, and it would then all pass).

Regardless, and after validating with some other models as well, 2098 seems to be the magic number.

I'll let this test run, so you see can see it fail first, before tweaking the number.

### DIFF
--- a/src/YesSql.Provider.SqlServer/SqlServerDialect.cs
+++ b/src/YesSql.Provider.SqlServer/SqlServerDialect.cs
@@ -243,6 +243,6 @@ namespace YesSql.Provider.SqlServer
 
         public override bool SupportsIfExistsBeforeTableName => true;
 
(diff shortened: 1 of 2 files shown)


sebastienros:
I couldn't understand why 2100 wasn't working and I had to set 2099. I presume same here, but it might just be delaying the next issue. I tried to update to 2000 but apparently your pr is protected or something. So let's see with that for now.

PR #345, opened 2021-02-09
support for "now" function which returns current DateTime in Utc

lampersky:


src/YesSql.Provider.PostgreSql/PostgreSqlDialect.cs
```diff
@@ -93,6 +93,7 @@ public PostgreSqlDialect()
             Methods.Add("day", new TemplateFunction("extract(day from {0})"));
             Methods.Add("month", new TemplateFunction("extract(month from {0})"));
             Methods.Add("year", new TemplateFunction("extract(year from {0})"));
+            Methods.Add("now", new TemplateFunction("now() at time zone 'utc'"));
```
Skrypt, line 96:
```suggestion
            Methods.Add("now", new TemplateFunction("(now() at time zone 'utc')"));
```

Skrypt: Using (now() AT TIME ZONE 'utc') would make this more bulletproof but I think it's not necessary. Just updated syntax since it should be upper case (just by convention).

PR #349, opened 2021-03-01
Added SessionReadOnly

jersiovic:
This PR provides a new interface ISessionReadonly that allows to create sessions only for read operation. It doesn't use transactions and it doesn't cache entities at identity map. 
It is useful for helping devs to reduce the number of locks when they use YesSql and for  reducing the probability of interlocks as it is explained in this issue https://github.com/OrchardCMS/OrchardCore/issues/8705

### DIFF
--- a/src/YesSql.Abstractions/ISession.cs
+++ b/src/YesSql.Abstractions/ISession.cs
@@ -10,7 +10,7 @@ namespace YesSql
     /// <summary>
     /// Represents a connection to the document store.
     /// </summary>
-    public interface ISession : IDisposable, IAsyncDisposable
+    public interface ISession : ISessionReadOnly, IDisposable, IAsyncDisposable
     {
         /// <summary>
         /// Saves a new or existing object to the store, and updates
@@ -51,27 +51,6 @@ namespace YesSql
         void Detach(object item, string collection = null);
 
         /// <summary>
-        /// Loads objects by id.
-        /// </summary>
-        /// <returns>A collection of objects in the same order they were defined.</returns>
-        Task<IEnumerable<T>> GetAsync<T>(int[] ids, string collection = null) where T : class;
-
-        /// <summary>
-        /// Creates a new <see cref="IQuery"/> object.
-        /// </summary>
-        IQuery Query(string collection = null);
-
-        /// <summary>
-        /// Executes a compiled query.
-        /// </summary>
-        /// <remarks>
-        /// A compiled query is an instance of a class implementing <see cref="ICompiledQuery{T}" />.
-        /// Compiled queries allow YesSql to cache the SQL statement that would be otherwise generated
-        /// on each invocation of the LINQ query. 
-        /// </remarks>
-        IQuery<T> ExecuteQuery<T>(ICompiledQuery<T> compiledQuery, string collection = null) where T : class;
-
-        /// <summary>
         /// Cancels any pending commands.
         /// </summary>
         void Cancel();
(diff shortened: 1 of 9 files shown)


sebastienros:
> It doesn't use transactions

I think I have read that it's important to do reads in transactions. Do you have another opinion?

> and it doesn't cache entities at identity map

Is that a performance issue?

sebastienros:
Using a custom interface is a solution, but maybe there are other options to brainstorm.
Like what if there was a mode where it would not create a transaction if there are no updates. Wouldn't that be better, since it would be transparent?
Or a mode where identitymap is not used.

Could even be options on the CreateSession() still, but return the same interface. Technically I don't think a custom interface is necessary since this is the same contract.

jersiovic:
Tests on this pr finally are fixed. Problem was I reproduced the interlock only for sqllite which was the only db I've testes the pr.  For sqlserver that test was not valid cause it throws an exception cause it doesn't accepts to have multiple transactions open for one connection. So, I've created only a test for sql server which creates two stores to simulate the scenario of two threads locking one to the other.

> > It doesn't use transactions
> 
> I think I have read that it's important to do reads in transactions. Do you have another opinion?
> 
You are right I didn't remember that point. I read somewhere the cost of opening a trasaction is important, and when you don't add a transaction every single query is created by the server with a transaction that is closed after it ends to ensure the atomicity of the operation. So, that's why are cheaper longer transactions.
Thinking on the problems we are facing I would add to that idea that it is important to measure the counterpart of having longer transactions, which is it is more likely they lock other operations. So the initial performance gain you would obtain with longer transactions in an scenario with low interactions between transactions will be lost and maybe it will be worse in an scenario with more locks interactions caused by those long transactions. So, it looks it is important to have tools for both scenarios.
In my case all the integration test create at least one user and request an access token to openid module, thats's why for me was enough to reinmplement UserStore, and openiddict stores using readonly sessions for read methods, and regular sessions for methods that store changes on db. By the momento no other changes are needed on my db. 

> > and it doesn't cache entities at identity map
> 
For my use case there is no problem on caching also for readonlysessions. I removed caching when I was writting the tests for this pr, cause I found weird that after another session writes sth at the db if you read from the db the data readed didn't have the change. But I understand that it could be valid depending on the scenario.
I found also strange to use the identitymap in a session that was not though for tracking changes. And because I saw that no matter if an entity is cached or not, the query to db is always done.
> Is that a performance issue?
I didn't though on that, I see that to use the identity map cache saves the session from  deserialize the entity, so to remove the cache has to have a penalty. I've meassured how much my 185 integration tests takes to execute. with and without identity map cache and times are almost the same. Maybe with test that handles bigger documents and queries on same request accessing same documents this difference is meaningful, but at least for my test it is not meaningful.

jersiovic:
> Using a custom interface is a solution, but maybe there are other options to brainstorm.
> Like what if there was a mode where it would not create a transaction if there are no updates. Wouldn't that be better, since it would be transparent?
> Or a mode where identitymap is not used.
> 
> Could even be options on the CreateSession() still, but return the same interface. Technically I don't think a custom interface is necessary since this is the same contract.

Yes, interface is only a helper to help users of the library which methods they should use for the mode they selected when the session was created. It is not a must have.
To have more modes on the creation of the session is a great help for supporting any scenario, the problem is it will lead to have multiple kinds of Session registered for IoC each one configured with one different mode.
What I would like is to use an approach like EF which allows to define the default mode in the creaton of the session but also to set the mode for one query, with extensiong like .AsNoTracking()
https://docs.microsoft.com/es-es/ef/core/querying/tracking
We could also add another one called .WithTransaction().

And for Session.GetAsync() queries we could provide optional params for doing the same

jersiovic:
In this way Orchard only will have a Session class registered for IoC that will be valid for using on every scenario we need

jersiovic:
Problem is a connection cannot execute queries without transactions when there is a transaction open, so it will need two connections per session, one for queries with transaction and the other for queries without transaction.

jersiovic:
Adding documentOID to indexes problems of this issue https://github.com/sebastienros/yessql/pull/349 dissapeared.
My integration tests are running on my own machine. I've the feeling that interlock problems would come back in production with more Azure instances running the app under heavy load. So, it could be worth to have sth like what we are talking about on YesSql to provide support for scenarios where indexes are not enough to avoid interlocks

Issue #343, opened 2021-02-02
Concurrency checking on null Version


jtkech:
Related to https://github.com/OrchardCMS/OrchardCore/issues/8470

After upgrading an application, if there is an existing document whose `Version` column is NULL, if we save this document using the checkConcurrency option, it fails because the generated command includes `where ... and [Version] = 0`.

As i understand an existing Version "can't" be equal to 0, but >= 1 or NULL, so when we update a document whose previous version has been mapped to 0, we could use `[Version] is null` or maybe `([Version] = 0 or [Version] is null)`.

Skrypt:
I will add that migrations might fail if this is not taken care of.

deanmarcussen:
I had a look at this, because I had some memory of fixing something related. Need to check more.

But @Skrypt can you look at the constraints on your `Document` table and see if you have a default(0) for the `Version` column. 

The create script should look include this...

`ALTER TABLE [dbo].[Document] ADD  DEFAULT ((0)) FOR [Version]`

If I remember rightly there was a short window where people were creating their own `Version` column because the migration failed, and I suspect anyone who created it manually didn't get the necessary constraint.

So it maybe that this impacts only on databases with the missing constraint.

Skrypt:
![image](https://user-images.githubusercontent.com/3228637/109189475-0b004700-7762-11eb-822c-6c52a30ec0cd.png)

Skrypt:
I think that the issue is not the default value. It's the fact that this column allows a null value.

deanmarcussen:
> That's probably why it didn't get migrated not allowing a null value.

yes, that's exactly why.

It maybe too late now, but it would still help to know if you had the constraint. Which I suspect you didn't, which is what's letting it be null.

The constraint should have forced it up to 0.

Skrypt:
I had no Version column it got added with the default(0) value after the table was created. So basically, when the column got added it was allowing a null value ; so for the records that we're there it set them as null. Because a default value always gets affected on the CREATE of a row record if you specify none. But it's not affecting a default value on an UPDATE.

PR #352, opened 2021-03-07
Handle null version column

deanmarcussen:
Fixes https://github.com/sebastienros/yessql/issues/343

On SQL Server @Skrypt is right, the Version column maybe `Null` which throws concurrency exceptions, when updating settings, or any of the other Documents.

It doesn't seem to happen on SQLite as the column gets set to 0.

Here when checking the `Version` column, if it is 1, i.e was the default of 0, then incremented to 1, I've added a null check on the version column.

When the Version value is over 1, we know the column cannot be null, so get the normal sql.

(it could just always use the null check, and would be no harm, if you wanted to keep the code simpler there)

Packed locally and tested with a OrchardCore SQL database with the `Version` column reset to null, and it works now.

Screenshot 2021-03-07 at 18 06 39


sebastienros:
Should we also solve it by not allowing NULL values on the column and forcing 0 as a default value? At least for new tables, and keeping this code to handle old tables.

deanmarcussen:
Make sense.

I think the original migration was nullable, to allow inserting the column, so I've left that in, but for new tables, not null.

Issue #218, opened 2019-11-21
Automatically generate sql indexes for DocumentId foreign keys


carlwoodhouse:
As the title says ? thoughts ? on large datasets and when you have many index tables it massively increases publishing performance (due to lots of delete where documentid = x, update where documentid = x) on index tables

PR #353, opened 2021-03-08
Adding indexes for index primary keys

sebastienros:


sebastienros:
Please check @deanmarcussen @jtkech

jtkech:
Okay, i will do

deanmarcussen:
Fixes https://github.com/sebastienros/yessql/issues/218

src/YesSql.Core/Sql/SchemaBuilder.cs
```diff
@@ -56,11 +56,16 @@ public ISchemaBuilder CreateMapIndexTable(Type indexType, Action<ICreateTableCom
...
                     .Column<int>("Id", column => column.PrimaryKey().Identity().NotNull())
-                    .Column<int>("DocumentId");
+                    .Column<int>("DocumentId")
+                    ;
 
                 table(createTable);
                 Execute(_commandInterpreter.CreateSql(createTable));
 
+                AlterTable(indexTable, table =>
+                    table.CreateIndex($"IDX_PK_{indexTable}", "Id")
```
deanmarcussen, line 66: I thought it was the foreign keys that weren't getting the indexes?

Primary key always gets one, but the `CreateForeignKey` below is the one that doesn't get an index on `DocumentId`, which is the value used for the join back to `Document`

sebastienros: I am glad I asked you to review, lame

sebastienros: Updated

sebastienros:
Apparently it also didn't define the PK on reduced index tables.

src/YesSql.Core/Sql/SchemaBuilder.cs
jtkech: Yes, as we removed the `CreateIndex()` for the `Identifiers` table through #350

src/YesSql.Core/Sql/SchemaBuilder.cs
```diff
@@ -84,7 +89,7 @@ public ISchemaBuilder CreateReduceIndexTable(Type indexType, Action<ICreateTable
                 var documentTable = TableNameConvention.GetDocumentTable(collection);
 
                 createTable
-                    .Column<int>("Id", column => column.Identity().NotNull())
+                    .Column<int>("Id", column => column.PrimaryKey().Identity().NotNull())
```
jtkech, line 92: Yes, i was looking at the code to see if here also you don't need an index on the `Id`, but was not sure because here you were not using `PrimaryKey()`, but then was still blocked on this line in `BaseCommandInterpreter .cs`

    var primaryKeys = command.TableCommands.OfType<CreateColumnCommand>()
        .Where(ccc => ccc.IsPrimaryKey && !ccc.IsIdentity).Select(ccc =>
            _dialect.QuoteForColumnName(ccc.ColumnName)).ToArray();

Where it seems that to be a primary key it needs to be `IsPrimaryKey` but not `IsIdentity`, that said looking in a sql table i could see them as primary keys.

sebastienros: Maybe in some dbs Indentity is a synonym of PK ... I will try to understand why it's filtering our Identity columns.
If they are not marked as PKs, then there would be no index.

sebastienros: Found it! The `IdentityColumnString` sql in the dialect will also set it as PK. That's why the PK statement is only needed to PKs that are not Identity. Adding some comments.

jtkech: Okay cool !!!

jtkech: So, would just using  `.Identity()` have been enough?

jtkech: Okay, just saw that you just updated it ;)

deanmarcussen:
Yeah looks good.

@jtkech we may have a couple more migrations to change if we made any with only the DocumentId, so there are no duplicated indexes

jtkech:
@deanmarcussen 

Yes, need to be checked when using `CreateMapIndexTable()`, hmm not sure that is good for us to add here this index with only `DocumentId`, we now always add indexes including it with other columns, so it will be an extra index and maybe not so useful as it only includes one column, but yes maybe still useful for people that don't create any index.

When using `CreateReduceIndexTable()` we never added any indexes as i was not sure for this kind of tables, so that's okay, hmm but does it means that would be good to add indexes also for these tables, if so same remark as above.

Finally, as i understand we can't add indexes on our side for the bridge table, so looks good to do it here including the `Id` and `DocumentId` columns, but how to include other columns of this bridge table, if useful.

I'm a little lost ;)

sebastienros:
> include other columns of this bridge table

There are no other columns, right? this index is useful since any query on a reduced index will always have a join on these two columns.

> we now always add indexes including it with other columns

This is an optimization based on usage. I think having the index on this PK still makes sense when you want to read all the indexed records for specific documents.

sebastienros:
Also these are indexes of INTEGER values, they are very efficient in terms of searching and memory consumption.

jtkech:
Found one for `AutoroutePartIndex`, not including the `Path` column as it is too long, so only `DocumentId`


            SchemaBuilder.AlterIndexTable<AutoroutePartIndex>(table => table
                .CreateIndex("IDX_AutoroutePartIndex_DocumentId", "DocumentId")
            );

So, it will be a duplicate ;)

jtkech:
@sebastienros 

> There are no other columns, right?

Oops okay, forgot it as i was writing, not obvious as one is the table name + `Id`

Then, okay for the other points as they are INTEGERs

jtkech:
So one question, can we add indexes in our side when using `CreateReduceIndexTable()`, we never added any indexes as i was not sure for this kind of tables. If so we could add indexes on these tables, there are only a few ones in OC.

And maybe remove the duplicate i mentioned above

Issue #354, opened 2021-03-10
**InvalidCastException** when using inheritance in my models within IndexProvider


lampersky:
when trying to register two indexes with models which are using inheritance I'm getting **InvalidCastException**.

here is my test case:
https://github.com/lampersky/yessql/tree/lampersky/index-provider-repro

```
Message: 
    System.InvalidCastException : Unable to cast object of type 'YesSql.Indexes.DescribeContext`1[YesSql.Tests.Models.AnotherPerson]' to type 'YesSql.Indexes.DescribeContext`1[YesSql.Tests.Models.Person]'.
  Stack Trace: 
    IIndexProvider.Describe(IDescriptor context) line 11
    Store.CreateDescriptors(Type target, String collection, IEnumerable`1 indexProviders) line 347
    Store.Describe(Type target, String collection) line 321
    Session.GetDescriptors(Type t, String collection) line 1283
    Session.MapNew(Document document, Object obj, String collection) line 1298
    Session.SaveEntityAsync(Object entity, String collection) line 296
    Session.FlushAsync() line 775
    Session.Dispose() line 650
    CoreTests.InheritedClassTest() line 5276
    --- End of stack trace from previous location ---
```

![image](https://user-images.githubusercontent.com/6403130/110609787-dba00000-818d-11eb-9526-cdbbfa7f4cf8.png)

lampersky:
I think **IsAssignableFrom** (inside CreateDescriptors at Store.cs) cause the problem.
Because of inheritance both **Person** and **AnotherPerson** are assignable from **Person**.

```
        internal IEnumerable<IndexDescriptor> CreateDescriptors(Type target, string collection, IEnumerable<IIndexProvider> indexProviders)
        {
            if (!DescriptorActivators.TryGetValue(target, out var activator))
            {
                activator = MakeDescriptorActivator(target);

                DescriptorActivators = DescriptorActivators.SetItem(target, activator);
            }

            var context = activator();

            foreach (var provider in indexProviders)
            {
                if (provider.ForType().IsAssignableFrom(target) &&
                    String.Equals(collection, provider.CollectionName, StringComparison.OrdinalIgnoreCase))
                {
                    provider.Describe(context);
                }
            }

            return context.Describe(new[] { target }).ToList();
        }
```

lampersky:
I was able to fix this problem with:

[...]

PR #356, opened 2021-03-11
index provider InvalidCastException fix

lampersky:
fixes: https://github.com/sebastienros/yessql/issues/354
related: https://github.com/OrchardCMS/OrchardCore/issues/8802

/cc @deanmarcussen


src/YesSql.Core/Store.cs
```diff
@@ -341,7 +341,7 @@ internal IEnumerable<IndexDescriptor> CreateDescriptors(Type target, string coll
 
             foreach (var provider in indexProviders)
             {
-                if (provider.ForType().IsAssignableFrom(target) &&
+                if (provider.ForType().Equals(target) &&
```
deanmarcussen, line 344: I'm not sure to be honest. 

I think from reading this that the `AnimalIndex` will never get written to if you're saving a `Dog`, which I think is what you want, but I'm not sure that it fits with all the other sub type arrangements within `YesSql`.

Have you had a read of the `ShouldQuerySubClasses()` tests and the behaviour that is expected of the actual document returned, rather than just what you can get in an index query?

lampersky: > I think from reading this that the AnimalIndex will never get written to if you're saving a Dog, which I think is what you want

exactly (animal and dog only have some parts in common, which doesn't mean to me writing one index should write another)

> Have you had a read of the ShouldQuerySubClasses() tests and the behaviour that is expected of the actual document returned, rather than just what you can get in an index query?

reading it now, so it looks like we have two different expected behaviors:
- one described by **ShouldQuerySubClasses**, where you can query base type and still expected to get documents of that base type and all others which inherited from base type,
- another one described by **ShouldRegisterIndexProviderWithModelInheritance**, where I tried to build two separated indexes (with inheritance) and expected that query over one index return some data and query over second index return other data.

That being said, I'm not sure if there is any chance to mix those two behaviors. Which means, that you can't build two **separate** indexes (one index derived from another).

I can live without this. Just need to duplicate some index fields.

lampersky: Now, I see that my example with **Animal** and **Dog** is not the best one, because you could expect saving dog will save animal too :-P

In my real world example, where I spotted this problem, I had one Index with some very common fields like PublishedUtc, CreatedUtc, and instead of duplicating those fields I wanted to derive from base index type.

PR #358, opened 2021-03-12
Fix Reduce Indexes in Collections

deanmarcussen:
Fixes https://github.com/sebastienros/yessql/issues/357

### DIFF
--- a/src/YesSql.Core/Session.cs
+++ b/src/YesSql.Core/Session.cs
@@ -1154,7 +1154,7 @@ namespace YesSql
                             }
                         }
 
-                        var dbIndex = await ReduceForAsync(descriptor, currentKey);
+                        var dbIndex = await ReduceForAsync(descriptor, currentKey, collection);
 
                         // if index present in db and new objects, reduce them
                         if (dbIndex != null && index != null)
@@ -1233,11 +1233,11 @@ namespace YesSql
             }
         }
 
-        private async Task<ReduceIndex> ReduceForAsync(IndexDescriptor descriptor, object currentKey)
+        private async Task<ReduceIndex> ReduceForAsync(IndexDescriptor descriptor, object currentKey, string collection)
         {
             await DemandAsync();
 
-            var name = _tablePrefix + descriptor.IndexType.Name;
+            var name = _tablePrefix + _store.Configuration.TableNameConvention.GetIndexTable(descriptor.IndexType, collection);
             var sql = "select * from " + _dialect.QuoteForTableName(name) + " where " + _dialect.QuoteForColumnName(descriptor.GroupKey.Name) + " = @currentKey";
 
             var index = await _connection.QueryAsync(descriptor.IndexType, sql, new { currentKey }, _transaction);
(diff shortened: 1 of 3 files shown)


test/YesSql.Tests/CoreTests.cs
```diff
@@ -218,14 +219,20 @@ public void CreateTables(IConfiguration configuration)
...
+                            "Col1"
                             );
 
                     builder.CreateMapIndexTable<PersonByNameCol>(column => column
                             .Column<string>(nameof(PersonByNameCol.Name)),
-                            "Collection1"
+                            "Col1"
                             );
 
+                   builder.CreateReduceIndexTable<PersonsByNameCol>(column => column
```
sebastienros, line 230: How does it work? Being able to use the same table for two types of indexes. Or are the names of reduce index generated differently?

sebastienros (approved):

test/YesSql.Tests/CoreTests.cs
deanmarcussen: By the `Col1` down the bottom. More explicitly it's `collection: Col1` so the table eventually becomes `Col1_PersonsByNameCol`

src/YesSql.Core/Session.cs
```diff
@@ -1233,11 +1233,11 @@ private async Task ReduceAsync()
             }
         }
 
-        private async Task<ReduceIndex> ReduceForAsync(IndexDescriptor descriptor, object currentKey)
+        private async Task<ReduceIndex> ReduceForAsync(IndexDescriptor descriptor, object currentKey, string collection)
         {
             await DemandAsync();
 
-            var name = _tablePrefix + descriptor.IndexType.Name;
+            var name = _tablePrefix + _store.Configuration.TableNameConvention.GetIndexTable(descriptor.IndexType, collection);
```
deanmarcussen, line 1240: so the only real change here is to pass the current collection name into the reduce query.

Everything else in the pr, is just reducing `Collection1` down to `Col1` so MySql doesn't have a fit about index name lengths

test/YesSql.Tests/CoreTests.cs
sebastienros: Actually I didn't notice the "s" in Persons, because there is also a PersonByName index

src/YesSql.Core/Session.cs
jtkech: @deanmarcussen 

We already reduced index names but it was for `PostgreSql` that truncates too long names, but i thought it was okay for `MySql` for which we had some problems but for index row sizes, see https://github.com/OrchardCMS/OrchardCore/issues/8649#issue-813113977

Good to know if index name lengths is also a problem with `MySql`, but i would just want to confirm it

deanmarcussen: Both MySql and Postgres are limited to 64 characters (table names, and indexes - basically everything).

The errors related to them just show up in different ways. (I think Postgres truncates, MySql throws, but it seems inconsistent).

If you give MySql no index name, it actually generates a unique one, so maybe something to look into.
But then we lose track of the index name. And am not sure if Postgres has the same feature

jtkech: Okay thanks

And i think that Postgres truncates before the 63 chars, that would explain why we had an issue on Postgres with 2 indexes starting with a same and quite long string, but not on MySql.

So yes, better to try it concretely, thanks for doing that

Issue #361, opened 2021-03-23
Reduce index names too long for MySql


deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/pull/8837#issuecomment-804700658

The new code to add the foreign keys for reduce indexes by default is proving too much for MySql (and I assume pgres).

This `MyTable_FK_OpenId_OpenIdAppByLogoutUriIndex_OpenId_Document_DocumentId`

from here https://github.com/sebastienros/yessql/blob/dd587b433c990dc18973957794aad88d868fd5f3/src/YesSql.Core/Sql/SchemaBuilder.cs#L110

(in `CreateReduceIndex`) is producing 70 chars with a small table prefix of `MyTable`

So many ways to fix, and none of them ideal ;)

They need to be unique per database.

But they don't really need to be especially descriptive, as we currently do, because we don't really need to refer back to them afterwards (unlike a table name)

So the dialect could simply truncate it, and tack a guid on the end, or it could be a hash. 

I think MySql handles this by generating it's own hash, when no name is provided. Couldn't say the same for postgres without looking.

Or there are other options like a custom table name convention for OpenId, which just has long class names.

c.f. nhibernate approach 
https://stackoverflow.com/questions/36766848/how-does-hibernate-generate-foreign-key-constraint-names

sebastienros:
I think using a hash of a deterministic name is fine, as we can still find it if necessary. Just keep a standard prefix like FK, IDX, ...

PR #362, opened 2021-03-24
Implement hash for index and foreign key names

deanmarcussen:
Fixes https://github.com/sebastienros/yessql/issues/361

Ended up with implementations on the dialect for both

`FormatKeyName` and `FormatIndexName` as they are for slightly different requirements.

The `HashHelper` went in utils, for want of a better place, to save having it in both postgres and mysql.

You'll recognise the design, just base 32. Kept them lowercase as it is more obvious that it's been generated.

but prefixed with `IDX_FK_` or `FK_`

Would have potentially been nice to keep the table prefix in, but became more complicated. So kept it simple

Screenshot 2021-03-24 at 16 07 57

### DIFF
--- a/src/YesSql.Core/Utils/HashHelper.cs
+++ b/src/YesSql.Core/Utils/HashHelper.cs
@@ -6,15 +6,20 @@ namespace YesSql.Utils
 {
     public static class HashHelper
     {
-        // Use of numeric characters is supported as long as the name does not start with a number.
+        // Use of numeric characters is supported but dialects must be prefixed with a valid character for the dialect.
         private static readonly string _encode32Chars = "0123456789abcdefghjkmnpqrstvwxyz";
 
         /// <summary>
-        /// Produces a 52 character hash.
+        /// Produces a 52 character hash with the supplied prefix prepended.
+        /// The length of the prefix plus the 52 char hash must be less than the dialects maximum length.
         /// </summary>
-        public static string HashName(string name)
+        public static string HashName(string prefix, string name)
         {
-            // Create returns a fips compatabile implementation.
+            if (string.IsNullOrEmpty(prefix))
+            {
+                throw new ArgumentNullException("A prefix is required");
+            }
+
             using (var hash = SHA256.Create())
             {
                 var bytes = Encoding.UTF8.GetBytes(name);
@@ -25,7 +30,7 @@ namespace YesSql.Utils
                 var long3 = BitConverter.ToInt64(hashed, 16);
                 var long4 = BitConverter.ToInt64(hashed, 24);
 
-                return ToBase32(long1, long2, long3, long4);
+                return prefix + ToBase32(long1, long2, long3, long4);
             }
         }
 
--- a/src/YesSql.Core/Utils/HashHelper.cs
+++ b/src/YesSql.Core/Utils/HashHelper.cs
@@ -1,0 +1,68 @@
+using System;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace YesSql.Utils
+{
+    public static class HashHelper
+    {
+        // Use of numeric characters is supported as long as the name does not start with a number.
+        private static readonly string _encode32Chars = "0123456789abcdefghjkmnpqrstvwxyz";
+
+        /// <summary>
+        /// Produces a 52 character hash.
+        /// </summary>
+        public static string HashName(string name)
+        {
+            // Create returns a fips compatabile implementation.
+            using (var hash = SHA256.Create())
+            {
+                var bytes = Encoding.UTF8.GetBytes(name);
+                var hashed = hash.ComputeHash(bytes);
+
+                var long1 = BitConverter.ToInt64(hashed, 0);
+                var long2 = BitConverter.ToInt64(hashed, 8);
+                var long3 = BitConverter.ToInt64(hashed, 16);
+                var long4 = BitConverter.ToInt64(hashed, 24);
+
+                return ToBase32(long1, long2, long3, long4);
+            }
+        }
+
+        private static string ToBase32(long long1, long long2, long long3, long long4)
+        {
+            var charBuffer = new char[52];
+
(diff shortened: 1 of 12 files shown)


deanmarcussen:
@jtkech can you have a look too?

sebastienros:
Why is the hash so scary ? ;)

src/YesSql.Provider.PostgreSql/PostgreSqlDialect.cs
```diff
@@ -149,6 +150,25 @@ public override string GetTypeName(DbType dbType, int? length, byte? precision,
             throw new Exception("DbType not found for: " + dbType);
         }
 
+        public override string FormatKeyName(string name)
+        {
+            if (name.Length >= 63)
```
sebastienros, line 155: What is the actual max length, so we can understand where the 63 comes from.

src/YesSql.Core/Utils/HashHelper.cs
```diff
@@ -0,0 +1,68 @@
+using System;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace YesSql.Utils
+{
+    public static class HashHelper
+    {
+        // Use of numeric characters is supported as long as the name does not start with a number.
```
sebastienros, line 9: Where is this enforced in the code?

src/YesSql.Core/Utils/HashHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+        private static readonly string _encode32Chars = "0123456789abcdefghjkmnpqrstvwxyz";
+
+        /// <summary>
+        /// Produces a 52 character hash.
+        /// </summary>
+        public static string HashName(string name)
+        {
+            // Create returns a fips compatabile implementation.
```
sebastienros, line 17: Are there multiple issues in the sentence?

src/YesSql.Core/Utils/HashHelper.cs
sebastienros: I think it's important to write a comment somewhere that we use a deterministic hash instead of a random number so we can find the name of the key later on. I was asking myself why it was not a random number, smaller, while looking at the PR.

jtkech:
@deanmarcussen 

> @jtkech can you have a look too?

Okay i will do this night

Yes i thought about a hash, in fact i 1rst thought about a generated id of 26 chars as we already use, but said to me that it was still too long to only replace the table name part, but yes i didn't think about using it for the whole key name (still suffixed by `FK_`, good idea!

Looks like we need a deterministic hash for both the index name and the key, but maybe still worth to take a look at our static `IdGenerator` just under the `OrchardCore.Abstractions` project, where i used an optimization about string allocations by using `String.Create()` where we pass to the delegate a `Span<char> buffer`.

src/YesSql.Provider.PostgreSql/PostgreSqlDialect.cs
deanmarcussen: Added links to the docs

src/YesSql.Core/Utils/HashHelper.cs
deanmarcussen: Added comments to the `ISqlDialect` methods that it is deterministic.

src/YesSql.Core/Utils/HashHelper.cs
deanmarcussen: So easiest just to remove this comment.

Here's the docs which specify fips compliance
https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha256.create?view=net-5.0#System_Security_Cryptography_SHA256_Create

deanmarcussen:
> Looks like we need a deterministic hash for both the index name and the key, but maybe still worth to take a look at our static `IdGenerator` just under the `OrchardCore.Abstractions` project, where i used an optimization about string allocations by using `String.Create()` where we pass to the delegate a `Span<char> buffer`.

@jtkech have you been dreaming about prs that you could do :) The `IDGenerator` not yet using `String.Create()` - but it would be great if it did.

I didn't here, because it's not going to be called much, plus is multi-targetted for 2.0 and 2.1, so I don't think `String.Create()` is supported on 2.0?

src/YesSql.Core/Utils/HashHelper.cs
deanmarcussen: > Where is this enforced in the code?

Changed the hash helper so it has to take a prefix. 

I didn't validate that the incoming prefix is a char, not numeric, that seems the responsibility of the dialect.

sebastienros (approved):

jtkech:
@deanmarcussen 

> @jtkech have you been dreaming about prs that you could do :) The IDGenerator not yet using String.Create() - but it would be great if it did.

I was talking about OrchardCore/OrchardCore.Abstractions/IdGenerator.cs

    public static class IdGenerator
    {
        // Some confusing chars are ignored: http://www.crockford.com/wrmg/base32.html
        private static readonly char[] _encode32Chars = "0123456789abcdefghjkmnpqrstvwxyz".ToCharArray();


        public static string GenerateId()
        {
            var guid = Guid.NewGuid().ToByteArray();

            return String.Create(26, guid, (buffer, guid) =>  // Here the String.Create ;)
            {
                var hs = BitConverter.ToInt64(guid, 0);
                var ls = BitConverter.ToInt64(guid, 8);


                char[] encode32Chars = _encode32Chars;
                buffer[0] = encode32Chars[(hs >> 60) & 31];
                ...
                ...
                buffer[25] = encode32Chars[ls & 31];
            });
        }
    }

deanmarcussen:
@jtkech 

Aha! code duplication instead.
Maybe we need to use that for this one which is used in most places `DefaultIdGenerator.cs`
It's in `Infrastructure`

jtkech:
Yes, i created this statis component to be used easily when i worked on DocumentManager ;)

PR #375, opened 2021-04-03
Fix dispose pattern

sebastienros:


src/YesSql.Core/Session.cs
```diff
@@ -900,7 +908,15 @@ public async ValueTask DisposeAsync()
 
             _disposed = true;
 
-            await CommitOrRollbackTransactionAsync();
+            try
```
hishamco, line 911: We can create an inner function instead of writing code twice

src/YesSql.Core/Session.cs
```diff
@@ -900,7 +908,15 @@ public async ValueTask DisposeAsync()
...
+            {
+                await CommitOrRollbackTransactionAsync();
+            }
+            catch
+            {
+                _transaction = null;
```
hishamco, line 917: I think this should go into a `finally` block for clean up

hishamco (changes requested):

src/YesSql.Core/Session.cs
sebastienros: No because the standard path does it if it doesn't fail

src/YesSql.Core/Session.cs
sebastienros: No, one is async the other on is not

PR #378, opened 2021-04-07
AllAsync and AnyAsync

deanmarcussen:
Replaces https://github.com/sebastienros/yessql/pull/369 

Easier to do another than deal with the merge conflicts.

`ValueTask` this time, since you already used it.

Adds some async overloads for .Any and .All

Bit of code duplication here, hopefully is ok :)

Allows to create async functions, that can then be passed into these methods.

### DIFF
--- a/src/YesSql.Core/Services/DefaultQuery.cs
+++ b/src/YesSql.Core/Services/DefaultQuery.cs
@@ -1291,11 +1291,33 @@ namespace YesSql.Services
                 return query;
             }
 
+            ValueTask<IQuery<T>> IQuery<T>.AnyAsync(params Func<IQuery<T>, ValueTask<IQuery<T>>>[] predicates)
+            {
+                // Scope the currentPredicate so multiple calls will not act on the new predicate.
+                var currentPredicate = _query._queryState._currentPredicate;
+                var query = ComposeQueryAsync(predicates, new OrNode());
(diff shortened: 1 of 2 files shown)


sebastienros (approved):

src/YesSql.Core/Services/DefaultQuery.cs
```diff
@@ -1302,6 +1313,17 @@ IQuery<T> IQuery<T>.All(params Func<IQuery<T>, IQuery<T>>[] predicates)
                 return query;
             }
 
+            ValueTask<IQuery<T>> IQuery<T>.AllAsync(params Func<IQuery<T>, ValueTask<IQuery<T>>>[] predicates)
+            {
+                // Scope the currentPredicate so multiple calls will not act on the new predicate.
+                var currentPredicate = _query._queryState._currentPredicate;
+                var query = ComposeQueryAsync(predicates, new AndNode());
+                // Return the currentPredicate to it's previous value, so another method call will act on the previous predicate.
+                _query._queryState._currentPredicate = currentPredicate;
```
jtkech, line 1322: @deanmarcussen 

Just after a quick look so not sure, If `ComposeQueryAsync()` is not awaited and mutates `_currentPredicate`, it may mutate it after it is retrieved line 1322

deanmarcussen: You might be right there. I will check tomorrow cheers :)

PR #382, opened 2021-04-22
Filter parsing

deanmarcussen:
Still a little bit to do here, but have packed it locally, and tested with OC.

Decided in the end, that it suits being couple of new projects in YesSql, so hopefully you agree.

All the tests are against YesSql.


sebastienros (approved):
After that you could add a new page in the Wiki showing a sample, and the rules of the grammar.

PR #384, opened 2021-05-09
Clear query order when adding new order statements

deanmarcussen:
When building queries through lots of abstraction interfaces, it's useful sometimes to be able to provide a default `OrderBy` and then later set another order statement.

This makes this possible, by clearing the statements if another is added later (otherwise it just gets confused when building the sql)


sebastienros:
Makes sense since there is a `ThenOrderBy`

PR #386, opened 2021-05-13
Increase mysql column sizes

deanmarcussen:
Fixes https://github.com/sebastienros/yessql/issues/313


src/YesSql.Core/Sql/Schema/ColumnCommand.cs
```diff
@@ -47,7 +47,9 @@ public IColumnCommand WithLength(int? length)
 
         public IColumnCommand Unlimited()
         {
-            return WithLength(10000);
+            // This is the highest value the current dialects can support.
+            // The resulting column type is dialect dependant.
+            return WithLength(4194304);
```
sebastienros, line 52:
```suggestion
            return WithLength(int.MaxValue);
```

deanmarcussen:
I checked this btw, and can't see any impact on MS SQL and PG

Issue #38, opened 2017-02-23
To set collection by configuration


jersiovic:
Currently a collection can be set only with an using statement. We talked about to offer a mechanism based on configuration to set which class uses which collection.

sebastienros:
More on that:
- Using an attribute on classes
- Using a metadata object in the configuration object

I don't like the using pattern, I must have had some reasons though, I assume I will figure it out.

sebastienros:
Also:
- Specify collection name in session methods (Query ...)

jersiovic:
My project it is been usin my own customized version of yessql with this PR https://github.com/sebastienros/yessql/pull/241
Based on that and on some wrappers at orchard side for Session and Query classes (CollectionPerTypeSession and CollectionPerTypeQuery) I was able to configure on which collection has to be stored an entity based on its class type. No matter if the code calling to `session.Save` is written on on my custom modules or in orchard core modules. This allowed me to store users, roles or what ever I want on the collection I want it.

Now I'm trying to come back to orchard core and yessql sources to get all the improvements you have been working on all this time. But I want to keep working this featured I've described.

The only thing that prevents I accomplish it is the signature of this ISession method:
`public IQuery Query(string collection = null)`

Why? because this is the only method in the ISession interface where I don't know which is the class type of the entity or index which is going to be used, so I cannot guess which is the collection for the query.
In my custom orchard and yessql implementation I delayed the moment to decide which is the related collection to later.
So, this was my CollectionPerTypeQuery class which is a wrapper for IQuery:>
```
public class CollectionQuery : IQuery
    {
        private readonly IQuery _query;
        private readonly YesSQLCollectionNameService _yesSQLCollectionNameService;

        public CollectionQuery(IQuery query, YesSQLCollectionNameService yesSQLCollectionNameService)
        {
            _query = query;
            _yesSQLCollectionNameService = yesSQLCollectionNameService;
        }
        public IQuery<object> Any()
        {
            return _query.Any();
        }

        public IQuery<T> For<T>(bool filterType = true) where T : class
        {
            var collectionName = _yesSQLCollectionNameService.GetCollectionNameForType(typeof(T));
            if (collectionName == null)
            {
                return _query.For<T>(filterType);
            }
            using (new NamedCollection(collectionName))
            {
                return _query.For<T>(filterType);
            }
        }

        IQueryIndex<T> IQuery.ForIndex<T>()
        {
[...]

PR #389, opened 2021-05-17
Added collection param to IQuery.For and IQuery.ForIndex method

jersiovic:
This PR tries to fix problem extending YesSql functionality at custom Orchard Core module described at this comment https://github.com/sebastienros/yessql/issues/38#issuecomment-842378167

After the change in this PR  a third party can provide  ISession, IQuery and ISchemaBuilder implementations which automatically choose the collection based on the type of the object to store/load instead of force the consumer to decide the collection.
In Orchard use an alternate ISession and IQuery like that allows to change the collection to store/load any type defined in own modules or in modules of the orchard framework


jersiovic:
Hi @sebastienros  any feedback related to this PR ?

src/YesSql.Core/Services/DefaultQuery.cs
```diff
@@ -30,7 +30,7 @@ public QueryState(ISqlBuilder sqlBuilder, IStore store, string collection)
 
         public string _bindingName = "a1";
         public Dictionary<string, List<Type>> _bindings = new Dictionary<string, List<Type>>();
-        public readonly string _documentTable;
+        public string _documentTable;
```
sebastienros, line 33: Maybe remove it from there and only compute it when the query is executed?

test/YesSql.Tests/CoreTests.cs
```diff
@@ -4155,6 +4155,8 @@ public async Task ShouldQueryInnerSelectWithCollection()
                 Assert.Equal(0, await session.Query<Person, PersonByNameCol>(collection: "Col1").Where(x => x.Name.IsNotInAny<PersonByBothNamesCol>(y => y.Firstname)).CountAsync());
 
                 Assert.Equal(2, await session.Query("Col1").For<Person>().With<PersonByNameCol>().Where(x => x.Name.IsInAny<PersonByBothNamesCol>(y => y.Firstname)).CountAsync());
+                Assert.Equal(2, await session.Query().For<Person>(collection:"Col1").With<PersonByNameCol>().Where(x => x.Name.IsInAny<PersonByBothNamesCol>(y => y.Firstname)).CountAsync());
+                Assert.Equal(2, await session.Query("Col2").For<Person>(collection: "Col1").With<PersonByNameCol>().Where(x => x.Name.IsInAny<PersonByBothNamesCol>(y => y.Firstname)).CountAsync());
```
sebastienros, line 4159: Actually I am confused now. `Query` already has a constructor with a collection, so the `For` call should use this one. So I read the issue again and I see that you are trying to hack the state of the query. I'd prefer a solution where you can plug a custom "CollectionResolver" that could be configured instead. So the `Query()` constructor would either take a collection name, or a `CollectionResolver`, or nothing to use default Collectionresolver from the configuration. That might be even better for your design.

PR #391, opened 2021-05-19
Update comments on SaveChangesAsync

deanmarcussen:
Noticed this outdated comment when hovering over it this morning.

and VS code decided to migrate the tasks file

### DIFF
--- a/src/YesSql.Abstractions/ISession.cs
+++ b/src/YesSql.Abstractions/ISession.cs
@@ -87,12 +87,11 @@ namespace YesSql
         Task FlushAsync();
 
         /// <summary>
-        /// Flushes any changes, commits the transaction, and disposes it.
+        /// Flushes any changes, commits the transaction, and disposes the transaction.
         /// </summary>
         /// <remarks>
-        /// Sessions are automatically committed when disposed, however calling <see cref="SaveChangesAsync"/>
-        /// is recommended before the session is disposed to prevent it from being called on a non-async
-        /// code path.
+        /// Sessions are not automatically committed when disposed, and <see cref="SaveChangesAsync"/>
+        /// must per called before disposing the <see cref="ISession"/>
         /// </remarks>
         Task SaveChangesAsync();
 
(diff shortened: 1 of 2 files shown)


src/YesSql.Abstractions/ISession.cs
```diff
@@ -87,12 +87,11 @@ public interface ISession : IDisposable, IAsyncDisposable
...
         /// <summary>
-        /// Flushes any changes, commits the transaction, and disposes it.
+        /// Flushes any changes, commits the transaction, and disposes the transaction.
         /// </summary>
         /// <remarks>
-        /// Sessions are automatically committed when disposed, however calling <see cref="SaveChangesAsync"/>
-        /// is recommended before the session is disposed to prevent it from being called on a non-async
-        /// code path.
+        /// Sessions are not automatically committed when disposed, and <see cref="SaveChangesAsync"/>
+        /// must be called before disposing the <see cref="ISession"/>
```
jtkech, line 94: > Sessions are not automatically committed when disposed

@deanmarcussen are you sure ;)

It is commited on dispose but calling SaveChangesAsync before is recommended to be done async, unless SUPPORTS_ASYNC_TRANSACTIONS is set.

https://github.com/sebastienros/yessql/blob/6b8c3e36421e68d1b1be5ed8c3319562d5851597/src/YesSql.Core/Session.cs#L637-L651

deanmarcussen: @jtkech I think so, but had to check :).

That was the change of behaviour in v3, that it wouldn't commit on dispose, so that exceptions, do not cause a commit.

So yes `CommitOrRollbackTransaction` is called, but no ops on commiting if save has not been called.
https://github.com/sebastienros/yessql/blob/6b8c3e36421e68d1b1be5ed8c3319562d5851597/src/YesSql.Core/Session.cs#L879-L884

Correct me if I'm wrong

jtkech: @deanmarcussen 

Okay cool then, `|| !_save` this is the change I missed, thanks

PR #394, opened 2021-06-02
Consume any char

deanmarcussen:
Consumes any char that isn't a term.

Fixes some issues with international text values


sebastienros (approved):

sebastienros:
No tests?

PR #413, opened 2021-11-20
Microsoft.Data.SqlClient 4.0.0

agriffard:
https://www.nuget.org/packages/Microsoft.Data.SqlClient/4.0.0
https://github.com/dotnet/SqlClient/releases/tag/v4.0.0


sebastienros:
The 4.0 version has a huge breaking change, that it requires an encrypted connection by default, and will validate the certificate too. If the connection string doesn't have `Encrypt=false` it won't work in our tests. I will wait a little until merging this one, to let the new version find it's place among users first.

PR #416, opened 2021-12-03
Install all .NET in one step

sebastienros:


.github/workflows/build.yml
```diff
@@ -53,22 +53,13 @@ jobs:
           MSSQL_SA_PASSWORD: Password12!
 
     steps:
-    - uses: actions/checkout@v2
```
lahma, line 56: You need to keep the checkout

lahma:
Did I trigger a merge with a comment 🤯

sebastienros:
What happened here?

sebastienros:
> enabled auto-merge

sebastienros:
And now:

> MSBUILD : error MSB1003: Specify a project or solution file. The current working directory does not contain a project or solution file.

sebastienros:
Resolved

PR #418, opened 2021-12-22
Add Schema to SQL Server and PostgreSQL dialect

### LINKED ISSUE #379: Violation of PRIMARY KEY constraint 'PK__t4h_Docu__3214EC07B34ED974'. Cannot insert duplicate key in object 'dbo.t4h_Document'. The duplicate key value is (81)
This Issue was first reported on Orchard Core:
https://github.com/OrchardCMS/OrchardCore/issues/9088

### In short:
Trying to create new Users in Orchard Core did not work and created the error and Stack-Trace reported to the ticket above.
### Context:
1. Two Instances of Orchard Core were running at the same time on the same database, one of them in Visual Studio, one as a live preview.
2. `Identifiers` table seemed to have a value above `max id` of documents, but the ids causing this error were in the range below `max id`
3. After stopping both instances and just starting one, the system recovered after one insuccessful try.
4. Even after this one instance recovered, the inserted new ids were below `max id` and the `nextval` value in `identifiers` pointed to the same value as when the system did not work.
5. Obviously, as lower ids could be inserted, there are unused ids below `max id` in my `documents` table
### Hypothesis
1. There is a mechanism (intended or not), that recovers unused ids below `max id`, but the ids that one instances claims in the range below `max id` are not represented in the database and therefore not communicated to other instances.

### Version
In my Orchard Core Version 1.0.0.0 RC2 stable, the used YesSql still seems to be way below the current version.

Skrypt:
Fixes #379

A Schema is a logical collection of database objects for these 2.

## TODO

- [x] Create Schema on the database from SchemaBuilder.
- [x] Make unit tests pass with a custom schema name in PostgreSQL
- [x] Make unit tests pass with a custom schema name in SQL Server
- [x] Make sure everything is backward compatible

### DIFF
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -182,7 +182,8 @@ namespace YesSql.Provider
         }
         public abstract string GetDropIndexString(string indexName, string tableName);
         public abstract string QuoteForColumnName(string columnName);
-        public abstract string QuoteForTableName(string tableName, bool IncludeSchema = false);
+        public abstract string QuoteForTableName(string tableName);
+        public abstract string SchemaNamePrefix();
 
         public virtual string QuoteString => "\"";
         public virtual string DoubleQuoteString => "\"\"";
@@ -192,6 +193,8 @@ namespace YesSql.Provider
         public virtual string DefaultValuesInsert => "DEFAULT VALUES";
 
         public virtual bool PrefixIndex => false;
+
+        public virtual string Schema => String.Empty;
 
         public abstract byte DefaultDecimalPrecision { get; }
 
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -198,6 +198,8 @@ namespace YesSql.Provider
 
         public virtual string Schema => null;
 
+        public virtual string DefaultSchema => null;
+
         public abstract byte DefaultDecimalPrecision { get; }
 
         public abstract byte DefaultDecimalScale { get; }
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -296,7 +296,7 @@ namespace YesSql.Provider
             return select;
         }
 
-        private readonly Dictionary<Type, List<Func<object, object>>> _typeHandlers = new Dictionary<Type, List<Func<object, object>>>();
+        private readonly Dictionary<Type, List<Func<object, object>>> _typeHandlers = new();
 
         public void ResetTypeHandlers()
         {
@@ -311,6 +311,11 @@ namespace YesSql.Provider
             }
 
             handlers.Add(i => handler((T)i));
+        }
+
+        public virtual string GetCreateSchemaString(string schema)
+        {
+            return $"CREATE SCHEMA {QuoteForColumnName(schema)}";
         }
     }
 }
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -182,7 +182,7 @@ namespace YesSql.Provider
         }
         public abstract string GetDropIndexString(string indexName, string tableName);
         public abstract string QuoteForColumnName(string columnName);
-        public abstract string QuoteForTableName(string tableName);
+        public abstract string QuoteForTableName(string tableName, bool IncludeSchema = false);
 
         public virtual string QuoteString => "\"";
         public virtual string DoubleQuoteString => "\"\"";
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -185,7 +185,7 @@ namespace YesSql.Provider
         public abstract string GetDropIndexString(string indexName, string tableName);
         public abstract string QuoteForColumnName(string columnName);
         public abstract string QuoteForTableName(string tableName);
-        public abstract string SchemaNameQuotedPrefix();
+        public virtual string SchemaNameQuotedPrefix() => null;
 
         public virtual string QuoteString => "\"";
         public virtual string DoubleQuoteString => "\"\"";
@@ -196,7 +196,7 @@ namespace YesSql.Provider
 
         public virtual bool PrefixIndex => false;
 
-        public virtual string Schema => String.Empty;
+        public virtual string Schema => null;
 
         public abstract byte DefaultDecimalPrecision { get; }
 
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -143,6 +143,7 @@ namespace YesSql.Provider
                 .Append(String.Join(", ", srcColumns))
 #endif
                 .Append(") references ")
+                .Append(SchemaNameQuotedPrefix())
                 .Append(destTable);
 
             if (!primaryKey)
@@ -171,6 +172,7 @@ namespace YesSql.Provider
                 sb.Append("if exists ");
             }
 
+            sb.Append(SchemaNameQuotedPrefix());
             sb.Append(QuoteForTableName(name)).Append(CascadeConstraintsString);
 
             if (SupportsIfExistsAfterTableName)
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -183,7 +183,7 @@ namespace YesSql.Provider
         public abstract string GetDropIndexString(string indexName, string tableName);
         public abstract string QuoteForColumnName(string columnName);
         public abstract string QuoteForTableName(string tableName);
-        public abstract string SchemaNamePrefix();
+        public abstract string SchemaNameQuotedPrefix();
 
         public virtual string QuoteString => "\"";
         public virtual string DoubleQuoteString => "\"\"";
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -185,6 +185,7 @@ namespace YesSql.Provider
         public abstract string GetDropIndexString(string indexName, string tableName);
         public abstract string QuoteForColumnName(string columnName);
         public abstract string QuoteForTableName(string tableName);
+        public abstract string QuoteForAliasName(string aliasName);
         public virtual string SchemaNameQuotedPrefix() => null;
 
         public virtual string QuoteString => "\"";
--- a/src/YesSql.Core/Provider/BaseDialect.cs
+++ b/src/YesSql.Core/Provider/BaseDialect.cs
@@ -12,7 +12,7 @@ namespace YesSql.Provider
     /// </summary>
     public abstract class BaseDialect : ISqlDialect
     {
-        public readonly Dictionary<string, ISqlFunction> Methods = new Dictionary<string, ISqlFunction>(StringComparer.OrdinalIgnoreCase);
+        public readonly Dictionary<string, ISqlFunction> Methods = new(StringComparer.OrdinalIgnoreCase);
 
         protected static Dictionary<Type, DbType> _propertyTypes;
 
@@ -125,7 +125,7 @@ namespace YesSql.Provider
         public virtual string FormatKeyName(string name) => name;
         public virtual string FormatIndexName(string name) => name;
 
-        public virtual string GetAddForeignKeyConstraintString(string name, string[] srcColumns, string destTable, string[] destColumns, bool primaryKey)
+        public virtual string GetAddForeignKeyConstraintString(string name, string[] srcColumns, string destQuotedTable, string[] destColumns, bool primaryKey)
         {
             var res = new StringBuilder(200);
 
@@ -143,8 +143,7 @@ namespace YesSql.Provider
                 .Append(String.Join(", ", srcColumns))
 #endif
                 .Append(") references ")
-                .Append(SchemaNameQuotedPrefix())
-                .Append(destTable);
+                .Append(destQuotedTable);
 
             if (!primaryKey)
             {
@@ -164,7 +163,7 @@ namespace YesSql.Provider
         public virtual bool SupportsIfExistsBeforeTableName => false;
         public virtual string CascadeConstraintsString => String.Empty;
         public virtual bool SupportsIfExistsAfterTableName => false;
-        public virtual string GetDropTableString(string name)
+        public virtual string GetDropTableString(string tableName, string schema)
         {
             var sb = new StringBuilder("drop table ");
             if (SupportsIfExistsBeforeTableName)
@@ -172,8 +171,7 @@ namespace YesSql.Provider
                 sb.Append("if exists ");
             }
 
-            sb.Append(SchemaNameQuotedPrefix());
-            sb.Append(QuoteForTableName(name)).Append(CascadeConstraintsString);
+            sb.Append(QuoteForTableName(tableName, schema)).Append(CascadeConstraintsString);
 
             if (SupportsIfExistsAfterTableName)
             {
@@ -182,11 +180,10 @@ namespace YesSql.Provider
 
             return sb.ToString();
         }
-        public abstract string GetDropIndexString(string indexName, string tableName);
+        public abstract string GetDropIndexString(string indexName, string tableName, string schema);
         public abstract string QuoteForColumnName(string columnName);
-        public abstract string QuoteForTableName(string tableName);
+        public abstract string QuoteForTableName(string tableName, string schema);
         public abstract string QuoteForAliasName(string aliasName);
-        public virtual string SchemaNameQuotedPrefix() => null;
 
         public virtual string QuoteString => "\"";
         public virtual string DoubleQuoteString => "\"\"";
@@ -196,8 +193,6 @@ namespace YesSql.Provider
         public virtual string DefaultValuesInsert => "DEFAULT VALUES";
 
         public virtual bool PrefixIndex => false;
-
-        public virtual string Schema => null;
 
         public virtual string DefaultSchema => null;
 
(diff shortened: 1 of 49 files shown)


src/YesSql.Provider.PostgreSql/PostgreSqlDialect.cs
```diff
@@ -113,6 +113,7 @@ public PostgreSqlDialect()
         public override string RandomOrderByClause => "random()";
         public override bool SupportsIfExistsBeforeTableName => true;
         public override bool PrefixIndex => true;
+        public string DefaultSchema => "public";
```
sebastienros, line 116:
```suggestion
        public string Schema { get; set; }
        public const string DefaultSchema = "public";
```

src/YesSql.Provider.PostgreSql/PostgreSqlDialect.cs
```diff
@@ -220,7 +221,7 @@ public override string QuoteForColumnName(string columnName)
 
         public override string QuoteForTableName(string tableName)
         {
-            return QuoteString + tableName + QuoteString;
+            return QuoteString + DefaultSchema + "." + tableName + QuoteString;
```
sebastienros, line 224:
```suggestion
            return QuoteString + Schema ?? DefaultSchema + QuoteString + "." + QuoteString + tableName + QuoteString;
```

src/YesSql.Provider.SqlServer/SqlServerDialect.cs
```diff
@@ -109,6 +109,8 @@ public SqlServerDialect()
 
         public override byte DefaultDecimalScale => 5;
 
+        public string DefaultSchema => "dbo";
```
sebastienros, line 112:
```suggestion
        public string Schema { get; set; }
        public const string DefaultSchema = "dbo";
```

src/YesSql.Provider.SqlServer/SqlServerDialect.cs
```diff
@@ -208,7 +210,7 @@ public override string QuoteForColumnName(string columnName)
 
         public override string QuoteForTableName(string tableName)
         {
-            return "[" + tableName + "]";
+            return "[" + DefaultSchema + "." + tableName + "]";
```
sebastienros, line 213:
```suggestion
            return "[" + Schema ?? DefaultSchema + "].[" + tableName + "]";
```

Skrypt:
Will require some more thinking. There are some issues here and there because for example we can't do :  

`SELECT * FROM dbo.Table AS dbo.Table2`

So, I'm changing _dialect.QuoteForTableName(value) for _dialect.QuoteForColumnName(value) in some places so that it works but it might break for other types of databases.

image

Skrypt:
image

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -48,12 +48,12 @@ public SqlBuilder(string tablePrefix, ISqlDialect dialect)
         public void Table(string table, string alias = null)
         {
             FromSegments.Clear();
-            FromSegments.Add(_dialect.QuoteForTableName(_tablePrefix + table));
+            FromSegments.Add(_dialect.QuoteForColumnName(_tablePrefix + table));
```
sebastienros, line 51: This one looks wrong

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -93,14 +93,14 @@ public virtual void InnerJoin(string table, string onTable, string onColumn, str
             }
             
             JoinSegments.Add(" INNER JOIN ");
-            JoinSegments.Add(_dialect.QuoteForTableName(_tablePrefix + table));
+            JoinSegments.Add(_dialect.QuoteForColumnName(_tablePrefix + table));
```
sebastienros, line 96: Looks wrong

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -93,14 +93,14 @@ public virtual void InnerJoin(string table, string onTable, string onColumn, str
...
             if (!String.IsNullOrEmpty(alias))
             {
-                JoinSegments.AddRange(new[] { " AS ", _dialect.QuoteForTableName(alias) });
+                JoinSegments.AddRange(new[] { " AS ", _dialect.QuoteForColumnName(alias) });
             }
             JoinSegments.AddRange(new[] {
-                " ON ", _dialect.QuoteForTableName(onTable), ".", _dialect.QuoteForColumnName(onColumn),
-                " = ", _dialect.QuoteForTableName(toTable), ".", _dialect.QuoteForColumnName(toColumn)
+                " ON ", _dialect.QuoteForColumnName(onTable), ".", _dialect.QuoteForColumnName(onColumn),
+                " = ", _dialect.QuoteForColumnName(toTable), ".", _dialect.QuoteForColumnName(toColumn)
```
sebastienros, line 103: Looks wrong

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -160,7 +160,7 @@ public virtual string FormatColumn(string table, string column, bool isAlias = f
                 table = _tablePrefix + table;
             }
 
-            return _dialect.QuoteForTableName(table) + "." + column;
+            return _dialect.QuoteForColumnName(table) + "." + column;
```
sebastienros, line 163: Looks wrong

Skrypt:
Yes I know but see, it works still. So, while it is wrong it passes. But I need to find a better way, I agree.

sebastienros:
It works because it just ignores the schema in this case, like it was doing before the PR, but that wrong, the goal is to specify the schema any time the table name is used as a table name.

sebastienros:
I suggest to use docker to install Postres locally, I believe I added the command line to run in the unit test for that.

sebastienros:
Ideally the unit tests dialects should be initialized with a different schema, some tests would probably fail.

src/YesSql.Core/Sql/SqlBuilder.cs
Skrypt: Here the issue is that it passes there when we create a table alias.
So it does something like : 

`SELECT * FROM [dbo].[Table] AS [dbo].[Table1] INNER JOIN [dbo].[JoinedTable] AS [dbo].[JoinedTable1] ON [dbo].[JoinedTable1].[Column] = [dbo].[Table1].[Column]`

This is invalid SQL to use the schema prefix on table aliases.

Skrypt:
I just tested this on a local PostgreSQL docker instance. See the CI passes too.

sebastienros:
The AS statements don't need the schema, they can use column quoting.

Skrypt:
I did not analyze everything yet. Just a first pass to see if that would actually work. And it seems like it would. So that's at least a good sign.

src/YesSql.Provider.SqlServer/SqlServerDialect.cs
```diff
@@ -206,9 +209,17 @@ public override string QuoteForColumnName(string columnName)
             return "[" + columnName + "]";
         }
 
-        public override string QuoteForTableName(string tableName)
+        public override string QuoteForTableName(string tableName, bool IncludeSchema = false)
         {
-            return "[" + tableName + "]";
+            var result = "[" + tableName + "]";
```
sebastienros, line 214: How is it different from `QuoteForColumn` when `includeSchema` is false? I even question semantically. Meaning that it could be the same but we'd want to keep them with different methods, but here I don't think it's "a table without the schema" we want to render.
I will double check with your previous examples.

Skrypt: Yes, that's why I had used QuoteForColumn because there was no difference. But not appropriate.

Skrypt:
Made some changes hope you like it.

Skrypt:
I need to find where Dapper binds the Table name schema or modify the BaseCommandInterpreter "I think". Else, it doesn't use the schema at all when creating tables. It interprets it as the table name simply.

hishamco:
@Skrypt if you need any sort of help to test the PR in SQL Server or Postgres please ping me

Skrypt:
image

Skrypt:
Working now. Please test it. Happy new year 😉

Skrypt:
Well, almost 😄

src/YesSql.Core/Sql/SchemaBuilder.cs
```diff
@@ -36,7 +36,7 @@ private void Execute(IEnumerable<string> statements)
             foreach (var statement in statements)
             {
                 _logger.LogTrace(statement);
-                Connection.Execute(statement, null, Transaction);
+                var test = Connection.Execute(statement, null, Transaction);
```
sebastienros, line 39:
```suggestion
                Connection.Execute(statement, null, Transaction);
```

src/YesSql.Core/Sql/SchemaBuilder.cs
```diff
@@ -205,10 +205,11 @@ public ISchemaBuilder AlterTable(string name, Action<IAlterTableCommand> table)
                 table(alterTable);
                 Execute(_commandInterpreter.CreateSql(alterTable));
             }
-            catch
+            catch(Exception ex)
             {
                 if (ThrowOnError)
                 {
+                    _logger.LogError(ex.Message);
```
sebastienros, line 212:
```suggestion
            catch
            {
                if (ThrowOnError)
                {
```

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -79,7 +79,7 @@ public virtual void InnerJoin(string table, string onTable, string onColumn, str
             // Don't prefix if alias is used
             if (alias != onTable)
             {
-                onTable = _tablePrefix + onTable;
+                onTable = _dialect.SchemaNameQuotedPrefix() + _tablePrefix + onTable;
```
sebastienros, line 82: Can you check why there is no quoting on this table name?

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -91,18 +91,20 @@ public virtual void InnerJoin(string table, string onTable, string onColumn, str
...
+
             JoinSegments.Add(" INNER JOIN ");
-            JoinSegments.Add(_dialect.QuoteForTableName(_tablePrefix + table));
+            JoinSegments.Add(_dialect.SchemaNameQuotedPrefix() + _dialect.QuoteForTableName(_tablePrefix + table));
+
             if (!String.IsNullOrEmpty(alias))
             {
                 JoinSegments.AddRange(new[] { " AS ", _dialect.QuoteForTableName(alias) });
             }
+
```
sebastienros, line 102: No schema here?

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -158,9 +160,12 @@ public virtual string FormatColumn(string table, string column, bool isAlias = f
             if (!isAlias)
             {
                 table = _tablePrefix + table;
+                return _dialect.SchemaNameQuotedPrefix() + _dialect.QuoteForTableName(table) + "." + column;
+            }
+            else
+            {
+                return _dialect.QuoteForTableName(table) + "." + column;
```
sebastienros, line 167: No schema here?

src/YesSql.Provider.PostgreSql/PostgreSqlDialect.cs
```diff
@@ -83,8 +85,13 @@ static PostgreSqlDialect()
             };
         }
 
-        public PostgreSqlDialect()
+        public PostgreSqlDialect(string schema)
```
sebastienros, line 88: Should have a default `null` value

src/YesSql.Provider.SqlServer/SqlServerDialect.cs
```diff
@@ -84,8 +84,13 @@ static SqlServerDialect()
             };
         }
 
-        public SqlServerDialect()
+        public SqlServerDialect(string schema)
```
sebastienros, line 87: Default `null` value

sebastienros (commented):
Very few locations where the schema is not prefixing table names, but I want to see what you have to say about my comments.

src/YesSql.Provider.SqlServer/SqlServerDialect.cs
Skrypt: default is "public"

src/YesSql.Provider.PostgreSql/PostgreSqlDialect.cs
Skrypt: default is "public"

src/YesSql.Core/Sql/SqlBuilder.cs
Skrypt: Yes, there is probably an issue in here and on line 82. Will try to figure it out.

src/YesSql.Core/Sql/SqlBuilder.cs
Skrypt: It's an Alias so it doesn't take a Schema prefix. The real table names take a Schema name but once it is aliased it's not called with a schema.

Skrypt:
Ok, now I'm getting to the strange part where it executes the tests fine one by one but not when I execute them all.

Skrypt:
Maybe some tweaks here and there but mostly done and working 🚀

src/YesSql.Core/Provider/BaseDialect.cs
```diff
@@ -193,6 +196,8 @@ public virtual string GetDropTableString(string name)
 
         public virtual bool PrefixIndex => false;
 
+        public virtual string Schema => String.Empty;
```
Skrypt, line 199: null or String.Empty?

src/YesSql.Provider.MySql/MySqlDialect.cs
```diff
@@ -248,6 +248,9 @@ public override string QuoteForTableName(string tableName)
             return "`" + tableName + "`";
         }
 
+        // MySQL doesn't have this feature
+        public override string SchemaNameQuotedPrefix() => String.Empty;
```
Skrypt, line 252: Null or String.Empty?

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -79,30 +79,40 @@ public virtual void InnerJoin(string table, string onTable, string onColumn, str
             // Don't prefix if alias is used
```
Skrypt, line 79: As noted here I removed prefixes for alias so either I revert this change or I keep it and we create a migration in Orchard Core.

src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -79,30 +79,40 @@ public virtual void InnerJoin(string table, string onTable, string onColumn, str
...
+            }
+            else
             {
                 toTable = _tablePrefix + toTable;
             }
 
             if (!String.IsNullOrEmpty(toAlias))
             {
-                toTable = toAlias;
+                toTable = _dialect.QuoteForTableName(toAlias);
```
Skrypt, line 100: Here I needed to add the `QuoteForTableName(toAlias)` because it was previously done all the time on line 113. So Alias were using `QuoteForTableName()`. Also, it is used on Alias because we also do it on line 108 but I'm not sure it is necessary to add Quotes on an Alias name in an INNER JOIN query; We could also try to remove QuoteForTableName() on lines 100 and 108 to see if tests pass.

Skrypt: Seems like it's working perfectly fine without quotes.

Skrypt: Well, I talked too quickly. PostgreSQL doesn't like it.

Skrypt:
Here is the issue with PostgreSQL : 

```SQL
SELECT BabyYoda."tpDocument".* FROM BabyYoda."tpDocument" INNER JOIN BabyYoda."tpArticlesByDay_Document" AS ArticlesByDay_Document_a1 ON ArticlesByDay_Document_a1."DocumentId" = BabyYoda."tpDocument"."Id" INNER JOIN BabyYoda."tpArticlesByDay" AS ArticlesByDay_a1 ON ArticlesByDay_a1."Id" = ArticlesByDay_Document_a1."ArticlesByDayId" WHERE ("ArticlesByDay_a1"."DayOfYear" = @p0) limit 1 limit 1 limit 1
```

The `WHERE` clause is not using the same `ALIAS` name as to how it is declared in the `INNER JOIN`. So the issue with PostgreSQL is that you need to be consistent by using the same ALIAS name everywhere in the Query. So, it supports alias quotes but it is not necessary. The QuoteForAliasName here is kind of a patch because it forces using quotes on an ALIAS name whereas the WHERE clause should be smart enough to know that the table it refers to is actually an ALIAS and thus not add quotes if it was not added previously when doing AS.

lnaie:
This is something I noticed was missing. Is it gonna make it?

lnaie:
Come on, guys! :) The PRs are pilling up.

Skrypt:
You can use the preview feed if you need it now.

Skrypt:
Ok just analyzed this and the MyGet feed only has what is pushed on the main branch. PR's are not pushed on MyGet which makes sense. So you will need to get it from my fork and create the Nuget packages and use them by adding the local folder in your nuget.config file if you want to use it now.

Skrypt:
By the way @sebastienros the PR is ready. Even though some things that I noted could be refactored this PR is ready.

Skrypt:
This PR is now 9 months old. I stopped progress on this mainly because we never had time to review/merge this.
@sebastienros

hishamco:
@Skrypt if there's anything with this PR I can do, let me know

Skrypt:
Nothing much. We just need to have time.

sebastienros:
Refactored it so that dialects don't depend on the schema (will be useful for OC too) but is part of the configuration.

Tests are failing here with PG but worked on my machine, maybe an issue if the schema doesn't exist yet. Related to that I think the InitializeAsync method should create the schema too if it doesn't exist. Might be helpful. Users can still create it if they want beforehand (with custom arguments).

sebastienros:
Could be valuable to add a unit test for PG that doesn't set the Schema, to be sure that it works without (not even setting the default value 'public'). Done by creating a subclass of PgSqlTest.

Skrypt:
Are you sure that MySQL supports Schemas as PG and SQL Server do?

Skrypt:
Good work @sebastienros 😄

PR #427, opened 2022-01-26
Add MyGet preview build feed

Skrypt:
As suggested on a meeting to be able to test packages builds.
Might require creating the MyGet feed.


sebastienros:
There needs to be two different files for build/pr, since we only want the packages when the PR is merged in main, like here: https://github.com/sebastienros/jint/tree/main/.github/workflows

The big different is that pr.yml is only triggered on prs, and doesn't publish a package

Skrypt:
Ok, done.

PR #436, opened 2022-04-04
Support pre 2014 SQL SERVER version drop table

jardg:
Hi I got some errors while droping table in a SQL 2012 instance. Then I saw DROP TABLE IF EXISTS wasn't supported on that version.
unnamed (1)

With this pr we can drop tables in every SQL Server version implementing the GetDropTableString method for SQL Server provider for YesSql


sebastienros:
I don't want to support pre-2017 since I can't test it easily. I suggest you create your own local dialect by inheriting from SqlDialect and overriding this specific method. I updated your PR with the new changes that include the schema now.

PR #443, opened 2022-05-11
Add TableNameAttribute - fix #442

### LINKED ISSUE #442: Add TableNameAttribute to allow overriding the table name
Currently the [DefaultTableNameConvention](https://github.com/sebastienros/yessql/blob/1d04a007d893f0287ecc721f087d4f9764f9c90f/src/YesSql.Core/Services/DefaultTableNameConvention.cs) used the index name and there is no easy way to specify the table name if one does not want to follow the default naming convention. Also, the class `Configuration` does not use dependency injection to resolve the dependency which make replacing the implementation of `ITableNameConvention ` difficult.

IMO, we should add an attribute called something like `TableNameAttribute` where the user can override the table name. Something like this

```
public class TableNameAttribute : Attribute
{
    public string Name { get; } = String.Empty;

    public TableNameAttribute(string tableName)
    {
        if (String.IsNullOrEmpty(tableName))
        {
            throw new ArgumentNullException(nameof(tableName));
        }

        Name = tableName;
    }
}

public class TableNameConvention : ITableNameConvention
{
    public const string DocumentTable = "Document";

    public string GetIndexTable(Type type, string collection = null)
    {
        string tableName = type.Name;

        var attribute = Attribute.GetCustomAttribute(type, typeof(TableNameAttribute)) as TableNameAttribute;

        if (attribute != null && !String.IsNullOrEmpty(attribute.Name))
        {
            tableName = attribute.Name;
        }

        if (String.IsNullOrEmpty(collection))
        {
            return tableName;
        }

        return collection + "_" + tableName;
    }

    public string GetDocumentTable(string collection = null)
    {
        if (String.IsNullOrEmpty(collection))
        {
            return DocumentTable;
        }

        return collection + "_" + DocumentTable;
    }
}
```

CrestApps:
Add `TableNameAttribute` to allow overriding the table name.

Fixes #442

### DIFF
--- a/samples/YesSql.Samples.Performance/Properties/launchSettings.json
+++ b/samples/YesSql.Samples.Performance/Properties/launchSettings.json
@@ -1,0 +1,8 @@
+{
+  "profiles": {
+    "WSL": {
+      "commandName": "WSL2",
+      "distributionName": ""
+    }
+  }
+}
--- a/src/YesSql.Abstractions/TableNameAttribute.cs
+++ b/src/YesSql.Abstractions/TableNameAttribute.cs
@@ -1,11 +1,11 @@
-﻿using System;
+using System;
 
-namespace YesSql.Services
+namespace YesSql
 {
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct)]
     public class TableNameAttribute : Attribute
     {
-        public string Name { get; } = String.Empty;
+        public string Name { get; }
 
         public TableNameAttribute(string tableName)
         {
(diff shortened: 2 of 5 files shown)


sebastienros:
I prefer not to introduce attributes for model classes.
Can't you implement this with your custom `ITableNameConvention` that would do that?

That's exactly why I made it an interface.

CrestApps:
Thank you! I needed that for OrchardCore CMS project. I put a PR request in OrchardCore to allow one to override the implementation https://github.com/OrchardCMS/OrchardCore/pull/11713

Thank you

Issue #444, opened 2022-06-02
Detected concurrency problem with Query Gating


Jag84:
There's a problem while trying to save two changes on the same Document concurrently.

I've simplify the issue into two tests as you can see in this PR at my YesSQL fork. The fork is  created now for showing you the problem:

https://github.com/Jag84/yessql/pull/1

If you have Query Gating enabled and you try to update a document concurrently, things couldn't work as it is expected. It seems sometimes document stored contains changes from another not commited changes. It causes that a retry operation attemps to remove an element which failed due to concurrent exception and it fails cause the element doesn't exist (despite its original delete operation failed due to a concurrent exception).

If you disable Query Gating, everything works fine.

If you want to check it by yourself, you only need to clone and run both `ShouldUpdateConcurrentDocumentsWithQueryGating` and `ShouldUpdateConcurrentDocumentsWithoutQueryGating` tests.

Jag84:
I can solve the failure removing the element that doesn't exist returning a copy of the Document on `ProduceAsync`. The changed method would be something like this: 

```
  internal async Task<T> ProduceAsync<T, TState>(WorkerQueryKey key, Func<TState, Task<T>> work, TState state)
  {
      if (!Configuration.QueryGatingEnabled)
      {
          return await work(state);
      }

      object content = null;

      while (content == null)
      {
          // Is there any query already processing the ?
          if (!Workers.TryGetValue(key, out var result))
          {
              // Multiple threads can potentially reach this point which is fine
#if !NET451
              // c.f. https://blogs.msdn.microsoft.com/seteplia/2018/10/01/the-danger-of-taskcompletionsourcet-class/
              var tcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
#else
              var tcs = new TaskCompletionSource<object>();
#endif

              Workers.TryAdd(key, tcs.Task);

              try
              {
                  // The current worker is processed
                  content = await work(state);
              }
              catch
              {
                  // An exception occurred in the main worker, we broadcast the null value
                  content = null;
                  throw;
              }
              finally
              {
                  // Remove the worker task before setting the result.
                  // If the result is null, other threads would potentially
                  // acquire it otherwise.
                  Workers.TryRemove(key, out result);

                  // Notify all other awaiters to return the result
                  tcs.TrySetResult(content);
              }
          }
          else
          {
              // Another worker is already running, wait for it to finish and reuse the results.
[...]

PR #453, opened 2022-09-07
Concurrency problem

jersiovic:
This pr solves bug reported here https://github.com/sebastienros/yessql/issues/444

@sebastienros  IMO it is sth important to solve cause it causes concurrent read and save operations for the same document would end storing a document wiht unexpected data

### DIFF
--- a/src/YesSql.Core/Store.cs
+++ b/src/YesSql.Core/Store.cs
@@ -388,6 +388,7 @@ namespace YesSql
             }
 
             object content = null;
+            object clonedContent = null;
 
             while (content == null)
             {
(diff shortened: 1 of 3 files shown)


sebastienros (commented):
Good catch, I will check if the cloning could be done by deserializing the same source string instead of doing a back and forth serialization. I am also surprised by the check on the int.

jersiovic:
We are glad to help. Ok we will keep an eye on what you commit finally to merge it in our project.
As far as I know the int check was due to a try and failure strategy. Because some queries return an int  instead of a complex object, and int doesn't need serialization and in fact it fails with current serialization line at code.

PR #464, opened 2022-12-13
Enumerable filters project

deanmarcussen:
This introduces a `Enumerable` filters project which allows projecting an enumerable into a Filter Parsing arrangement, much as we can do with an `IQuery` currently.

Filtering on an `IEnumerable` is useful, for the case where we are using a single SQL document, i.e. like a `TemplatesDocument` where the items we want to filter on are contained in a dictionary, or list.

You may or may not want this in the YesSql project :) but I'm using it internally, so figured it was worth pushing up somewhere.

Note: There seems to be a slight bug in Partlot compilation which is causing the tests to fail randomly. I assume it's related to initialization, when initializing in parallel. Only applies to complied parsers.

I'll open an issue there to go with


sebastienros:
@copilot add a README.MD at the same level as the new project (not the tests) describing its goals and usages.

deanmarcussen:
100% @Skrypt

deanmarcussen:
ok, probably just due to a change in Parlot, I'll try and make time today to pull and do the same fix

deanmarcussen:
build fixed @Skrypt

PR #473, opened 2023-01-25
Add store state

MikeAlhayek:


sebastienros:
I was thinking about doing this in Orchard, not directly here. But I want to see if all test pass before thinking more about it.

MikeAlhayek:
@sebastienros yes the change will need to also be done in OC. But I think here we need to add `await _store.InitializeAsync();` in the Session to ensure in OC we always create connection using a pre-configured store.

sebastienros:
> yes the change will need to also be done in OC

I meant "no change in yessql". Not "also react in Orchard"

src/YesSql.Core/Store.cs
```diff
@@ -84,6 +87,11 @@ internal Store(IConfiguration configuration) : this()
 
         public async Task InitializeAsync()
         {
+            if (_state == StoreState.Initialized)
```
sebastienros, line 90: Should just be a `private bool _isInitialized` for now. No need for a new enum.

MikeAlhayek: sure. Done

jtkech:
Yes this is a pattern we are using in OC.

I would have named the method `EnsureInitialized()`.

Why using a enum for only 2 states, I would have used an `_initialized` bool.

But this is not a full lock, maybe use an async lock, e.g. a `SemaphoreSlim`.

MikeAlhayek:
@sebastienros @jtkech I think we may need to do a bit more here in the ISession. I think here we would need to ensure not only that the store is initialized, but every collection is initialized. in OC we have to also remove the loop call to `InitializeCollectionAsync(...)`

sebastienros:
Note that now the possibility to have multiple concurrent calls to `InitializeAsync()` is much higher, since while it is initializing other request can use the same store. When using `CreateAndInitializeAsync` it was called once, in an atomic way. We might want to block here too because that might break. So maybe add a semaphore on the initialization check.

sebastienros:
So it seems @jtkech have the same reflexes ;)

MikeAlhayek:
@sebastienros before I adventure off, thought on also passing collections to the store so we can initialize all collections?

sebastienros:
It's getting harder and harder. So maybe go back to the option which is to change it in Orchard only and invoke these initializations in an async context. For instance from a middleware when the DI has been registered. @jtkech is there an even we can plug to when this is done? Maybe like TenantActivatedAsync ?

In there we could then call InitializeAsync and InitializeCollectionsAsync

jtkech:
Hmm also just for info, just saw that `store.InitializeCollectionAsync()` open and use a new connection for each collection (at least one), and at the end call `IdGenerator.InitializeCollectionAsync()` which also open and use a new connection.

sebastienros:
@jtkech yeah this will never end. That's why I think this should be solved in OC.

jtkech:
@sebastienros just saw your comment, okay will look at it

src/YesSql.Core/Store.cs
```diff
@@ -82,8 +91,24 @@ internal Store(IConfiguration configuration) : this()
...
+        }
+
         public async Task InitializeAsync()
         {
+            if (_isInitialized)
+            {
+                return;
+            }
+
+            await _semaphore.WaitAsync();
```
jtkech, line 110: Then after good to double check `_isInitialized`.

MikeAlhayek:
@jtkech trying to solve the issue in OC directly. If needed, I'll update this PR. but if we can solve it in OC, maybe we don't need this.

src/YesSql.Core/Store.cs
```diff
@@ -82,8 +91,26 @@ internal Store(IConfiguration configuration) : this()
...
         public async Task InitializeAsync()
         {
+            await _semaphore.WaitAsync();
+
+            if (_isInitialized)
+            {
+                _semaphore.Release();
+
+                return;
+            }
```
jtkech, line 112: I meant something like this

            if (_isInitialized)
            {
                return;
            }

            await _semaphore.WaitAsync();
            try
            {
                if (_isInitialized)
                {
                    return;
                }

                ...
                ...
                _isInitialized = true;
            }
            finally
            {
                _semaphore.Release();
            }

MikeAlhayek: Done.

MikeAlhayek:
@jtkech trying to solve this directly in OC. Trying to find the proper event that should we hook into. I thought we can use `IModularTenantEvents` but it may not be the right event. Do we need a new event for this? 

in `ShellScope` does not call the events if `_serviceScopeOnly` is true.

```
internal async Task ActivateShellInternalAsync()
        {
            if (ShellContext.IsActivated)
            {
                return;
            }

            if (_serviceScopeOnly)
            {
                return;
            }

            // Try to acquire a lock before using a new scope, so that a next process gets the last committed data.
            (var locker, var locked) = await ShellContext.TryAcquireShellActivateLockAsync();
            if (!locked)
            {
                throw new TimeoutException($"Failed to acquire a lock before activating the tenant: {ShellContext.Settings.Name}");
            }

            await using var acquiredLock = locker;

            // The tenant gets activated here.
            if (!ShellContext.IsActivated)
            {
                await new ShellScope(ShellContext).UsingAsync(async scope =>
                {
                    var tenantEvents = scope.ServiceProvider.GetServices<IModularTenantEvents>();
                    foreach (var tenantEvent in tenantEvents)
                    {
                        await tenantEvent.ActivatingAsync();
                    }

                    foreach (var tenantEvent in tenantEvents.Reverse())
                    {
                        await tenantEvent.ActivatedAsync();
                    }
                }, activateShell: false);

                ShellContext.IsActivated = true;
            }
        }
```

jtkech:
> in ShellScope does not call the events if _serviceScopeOnly is true.

Yes I know, not sure ATM what is best to do, on OC or YesSql side, need to focus on it a little, will see.

src/YesSql.Core/Store.cs
```diff
@@ -82,33 +91,66 @@ internal Store(IConfiguration configuration) : this()
...
+            if (_isInitialized)
+            {
+                return;
+            }
 
-            if (!string.IsNullOrEmpty(Configuration.Schema))
+            try
             {
-                await using (var connection = Configuration.ConnectionFactory.CreateConnection())
+                await _semaphore.WaitAsync();
```
jtkech, line 112: Just before the try block, so that the semaphore is only released if it has been acquired

MikeAlhayek: Done

MikeAlhayek:
@jtkech @sebastienros I submitted a PR to attempt to solve this issue in OC directly https://github.com/OrchardCMS/OrchardCore/pull/13147

sebastienros:
That's exactly what I had in mind. Minus the change in the shell scope which I don't understand. So as long as it works well then that would be perfect.

We might even want to extend this pattern if we think the Tenants events are not the best because of this constraint (the scope thinggy change).

jtkech:
https://github.com/OrchardCMS/OrchardCore/pull/13147#issuecomment-1404445634

MikeAlhayek:
This was fixed in OrchardCore instead https://github.com/OrchardCMS/OrchardCore/pull/13147

Issue #477, opened 2023-03-09
SqlBuilder InnerJoin implementation does not match the interface parameter order


douwinga:
I ran into an issue while using the SqlBuilder with an InnerJoin that messed me up for a bit. 

The [interface ](https://github.com/sebastienros/yessql/blob/aa87d8f8edfdc9358127a90d49c282bf63308d50/src/YesSql.Abstractions/ISqlBuilder.cs#L15) has the order of parameters as `string table, string onTable, string onColumn, string toTable, string toColumn, string schema, string alias = null, string toAlias = null`. 

The [implementation of the interface](https://github.com/sebastienros/yessql/blob/aa87d8f8edfdc9358127a90d49c282bf63308d50/src/YesSql.Core/Sql/SqlBuilder.cs#L78) has the order of parameters as `string table, string onTable, string onColumn, string toTable, string schema, string toColumn, string alias = null, string toAlias = null`. 

The `shema` and `toColumn` parameters are swapped.

PR #478, opened 2023-03-10
Fixed parameter order of InnerJoin interface

douwinga:
Fix for #477 

I went with the non-breaking change way to fix this, but the parameter order seems odd now.

### DIFF
--- a/src/YesSql.Abstractions/ISqlBuilder.cs
+++ b/src/YesSql.Abstractions/ISqlBuilder.cs
@@ -12,7 +12,7 @@ namespace YesSql
         string FormatColumn(string table, string column, string schema, bool isAlias = false);
         string FormatTable(string table, string schema);
         string GetSelector();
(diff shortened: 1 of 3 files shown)


hishamco:
I think we need to change the implementation not the interface

douwinga:
I agree, I just wasn't sure if that was a good idea since anybody using this will have queries failing after updating. There wouldn't be any indication as to what changed and that they need to flip the parameters. However, if somebody else is using this, you would expect they would have ran into the same problem I did and have reported it. So it is probably a non issue.

hishamco:
> There wouldn't be any indication as to what changed and that they need to flip the parameters.

You are right, but may be we could log it into the release notes

sebastienros:
Please fix the implementation, users should be using the interface anyway.

douwinga:
It looks like I did fix the implementation with the commits from 7 months ago

sebastienros (approved):

PR #480, opened 2023-03-13
Advanced query index

### LINKED ISSUE #479: Select Index instead of document with All/Any support
As mentioned in #467 and #422, there are times that we need to query an index to retrieve some values, but the documents should be filtered with multiple indexes. Unfortunately this can't be done now, as the IQuery<T> returns documents (a large payload, which must be materialized in memory to read the specific info) or use the IQueryIndex<T> but without support for All/Any.

A solution is to introduce a ForAdvancedIndex<T>() method to IQuery, which would select from the Index table joined with the document table. The method should return an IQuery<T> instance, where then all the other methods work as the simple For<T> method. The method will only allow to select from Map Indexes.

MichaelPetrinolis:
Can select from the Index and supports All/Any

closes #479

### DIFF
--- a/src/YesSql.Core/Sql/SqlBuilder.cs
+++ b/src/YesSql.Core/Sql/SqlBuilder.cs
@@ -71,10 +71,10 @@ namespace YesSql.Sql
         public void Take(string take)
             => _count = take;
 
-        public virtual void Join(JoinType type, string table, string onTable, string onColumn, string toTable, string toColumn, string schema, string alias = null, string toAlias = null)
+        public virtual void Join(JoinType type, string table, string onTable, string onColumn, string toTable, string toColumn, string schema, string alias = null, string toAlias = null, bool onTableIsAlias = false)
         {
             // Don't prefix if alias is used
-            if (alias != onTable)
+            if (alias != onTable && !onTableIsAlias)
             {
                 onTable = _dialect.QuoteForTableName(onTable, schema);
             }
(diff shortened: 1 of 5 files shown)


src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -71,10 +71,10 @@ public void Skip(string skip)
         public void Take(string take)
             => _count = take;
 
-        public virtual void Join(JoinType type, string table, string onTable, string onColumn, string toTable, string toColumn, string schema, string alias = null, string toAlias = null)
+        public virtual void Join(JoinType type, string table, string onTable, string onColumn, string toTable, string toColumn, string schema, string alias = null, string toAlias = null, bool onTableIsAlias = false)
         {
             // Don't prefix if alias is used
-            if (alias != onTable)
+            if (alias != onTable && !onTableIsAlias)
```
sebastienros, line 77: Seeing this, I believe we don't need a new field and just to set the `alias` argument instead such that it knows that `onTable` is an alias.

hishamco:
@lampersky could you please fix the build and react to Seb comment, I think there several PRs stuck for a while

MichaelPetrinolis:
Hello @hishamco, I think that you wanted to mention me ;) I will try this week to fix the npgsql test failure.

hishamco:
Sorry I just messed up with another PR :)

Issue #483, opened 2023-03-30
The 'modify' keyword is not supported in postgres


hyzx86:
I don't understand what this code does. Why compare lengths?

https://github.com/sebastienros/yessql/blob/c0771eb555c67bf419859016cb3563a3540fe7c7/src/YesSql.Provider.PostgreSql/PostgreSqlCommandInterpreter.cs#L18-L22

https://github.com/sebastienros/yessql/issues/244#issuecomment-1489671512

hyzx86:
```C#
         SchemaBuilder.AlterIndexTable<AuditTrailEventIndex>(table => table
                        .AlterColumn(nameof(AuditTrailEventIndex.CorrelationId),
                            column => column.WithType(typeof(string), 255)),
                    collection: AuditTrailEvent.Collection
                );
```


```
Error while running migration version 0 for 'EasyOC.AuditTrail'. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> Npgsql.PostgresException (0x80004005): 42601: syntax error at or near "modify"

POSITION: 42
   at Npgsql.Internal.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|221_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
   at Npgsql.NpgsqlDataReader.NextResult()
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteNonQuery(Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteNonQuery()
   at Dapper.SqlMapper.ExecuteCommand(IDbConnection cnn, CommandDefinition& command, Action`2 paramReader) in /_/Dapper/SqlMapper.cs:line 2848
[...]

PR #485, opened 2023-04-05
fix pgsql alter column

hyzx86:
relate : https://github.com/sebastienros/yessql/issues/483


sebastienros:
Could you add a unit test that would fail without this fix? And even better if it's in the base unit test class such that it triggers the `IAlterColumnCommand` for all the databases.

PR #486, opened 2023-04-07
Fix alter column issue in Postgres

### LINKED ISSUE #483: The 'modify' keyword is not supported in postgres

hishamco:
Fixes #483


hishamco:
While I'm working on this I found two issues:

1- `modify` is incorrect as discussed on the related issue
2- Also, we miss `Type` when we change the column datatype, seems no one alter the column previously in Postgres :)

One more thing that confused me, I'm expecting to not specify the type during the alteration, if we don't the default `bytea` is added to the column which is wrong

hishamco:
Nah, SQLite test is failed!! I presume because it doesn't support alteration if I'm not wrong

MikeKry:
@hishamco 

I would also expect that it is failing because sqlite does not support alter column.. 

Can this be merged? It is vital to have possibility to change column size in postgres.

hishamco:
@sebastienros could you please reply on my [comment](https://github.com/sebastienros/yessql/pull/486#issuecomment-1500723187) so we can finalize this PR?

sebastienros (approved):

hishamco:
Ready to merge Seb

PR #490, opened 2023-07-24
fixes offset limit duplication fo sqlServer

### LINKED ISSUE #433: sqlbuilder generates multiple "limit" and "offset" clause when invoke ToSqlBuilder multiple times
I use postgres dialect sqlbuilder to generate customized sql. The .Skip() and .Take() is used in my sqlbuilder and found that a new "limit" and "offset" clause is generated as long as .ToSqlBuilder been invoked. If I invoke .ToSqlBuilder twice, there will be two paires of "limit" and "offset". If I invoke .ToSqlBuilder() three times. There will be three pares.

brunoAltinet:
this fixes when different skip/take is used on a same query

Fixes #433 

often a case when doing batching

### DIFF
--- a/src/YesSql.Provider.SqlServer/SqlServerDialect.cs
+++ b/src/YesSql.Provider.SqlServer/SqlServerDialect.cs
@@ -172,6 +172,7 @@ namespace YesSql.Provider.SqlServer
         public override void Page(ISqlBuilder sqlBuilder, string offset, string limit)
         {
             sqlBuilder.ClearTrail();
+
             if (offset != null)
             {
                 sqlBuilder.Trail(" OFFSET ");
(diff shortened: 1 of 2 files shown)


src/YesSql.Provider.SqlServer/SqlServerDialect.cs
```diff
@@ -171,6 +171,7 @@ public override string GetTypeName(DbType dbType, int? length, byte? precision,
 
         public override void Page(ISqlBuilder sqlBuilder, string offset, string limit)
         {
+            sqlBuilder.ClearTrail();
```
hishamco, line 174:
```suggestion
            sqlBuilder.ClearTrail();

```

test/YesSql.Tests/CoreTests.cs
```diff
@@ -2914,6 +2914,9 @@ public async Task ShouldPageResults()
                 {
                     Assert.Equal(1, String.Compare(ordered[i].SomeName, ordered[i - 1].SomeName));
                 }
+
+                //query should discard previous skip/take
+                Assert.Equal(5, (await query.Skip(10).Take(5).ListAsync()).Count());
```
hishamco, line 2919: Can we create a new test for this?

hishamco (changes requested):

sebastienros:
I reverted it since it was failing the PostgreSQL tests.

PR #501, opened 2023-10-11
Add Schema to IndexCommand CompoundKey

jtkech:
Related to https://github.com/OrchardCMS/OrchardCore/issues/14036#issuecomment-1661352012


MikeAlhayek:
@sebastienros if you approve this, we should also include it in 3.3.1 patch.

sebastienros:
Tried to use a `record` instead, then got carried away since it's only available since net6.0, at least for the supported frameworks.

PR #506, opened 2023-11-08
Update dependencies

MikeAlhayek:
1. Added `Directory.Packages.props` file to manage the dependencies.
2. Removed deprecated package.
3. Update all packages to use the latest package


samples/YesSql.Bench/YesSql.Bench.csproj
```diff
@@ -1,23 +1,23 @@
...
-  </PropertyGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\..\src\YesSql.Core\YesSql.Core.csproj" />
-    <ProjectReference Include="..\..\src\YesSql.Provider.SqlServer\YesSql.Provider.SqlServer.csproj" />
-  </ItemGroup>
-</Project>
\ No newline at end of file
+    <PropertyGroup>
+        <TargetFrameworks>net7.0</TargetFrameworks>
+        <DebugType>portable</DebugType>
```
jtkech, line 4: Just for info, I have no preference but at least in OC we use 2 spaces indentations e.g. for .csproj and .targets files.

MikeAlhayek: @jtkech VS auto format on save for me. So usually, I leave formating for VS so I don't have to manually adjust them. 

But, it is weird that the build is failing 

> error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: Microsoft.SourceLink.GitHub.


I am only using `PackageVersion` in the `Directory.Packages.props` file and in the projects I am using `PackageReference` with no version.

```
<PackageReference Include="Dapper.StrongName" />
```


I also, can't run the project locally for some reason (but it is the case even before my changes to the branch)

image

jtkech: Try to rename your file to `Directory.Build.props`.

Or set `ManagePackageVersionsCentrally` to true

MikeAlhayek: @jtkech the `Directory.Packages.props` is the right name which already contains `ManagePackageVersionsCentrally`. found and fixed the `error NU1008` issue. 

Now something is wrong with package framework dependency

image

PR #507, opened 2023-11-08
Cleanup

MikeAlhayek:


src/YesSql.Core/Services/DefaultQuery.cs
```diff
@@ -1240,17 +1240,17 @@ protected async Task<T> FirstOrDefaultImpl()
...
 
                             return state.Connection.QueryAsync<Document>(state.Sql, state.Query._queryState._sqlBuilder.Parameters, state.Transaction);
-                        }, new { Query = _query, Sql = sql, Connection = connection, Transaction = transaction })).ToArray();
+                        }, new { Query = _query, Sql = sql, Connection = connection, Transaction = transaction });
+
+                        // Clone documents returned from ProduceAsync as they might be shared across sessions
+                        var clonedDocuments = documents.Select(x => x.Clone()).ToList();
 
-                        if (documents.Length == 0)
+                        if (clonedDocuments.Count == 0)
```
sebastienros, line 1248: What's the reason to move it down?

MikeAlhayek: @sebastienros we need to check if documents contains anything then convert to to list. Moving it up will allow us to use Count instead of `Any()` to check of there were any records.

PR #508, opened 2023-11-13
Add Async method to SchemaBuilder to prevent possible thread starvation

### LINKED ISSUE #509: Add Async methods to the SchemaBuilder
Add Async method to SchemaBuilder to prevent possible thread starvation

MikeAlhayek:
Fix #509 

Note, there was a bug in the dispose logic of the Session class. The method `CommitOrRollbackTransactionAsync` has slightly different logic than `CommitOrRollbackTransaction`. I believe the logic in `CommitOrRollbackTransaction` is better and used.


src/YesSql.Core/Sql/SchemaBuilder.cs
```diff
@@ -69,11 +51,11 @@ public ISchemaBuilder CreateMapIndexTable(Type indexType, Action<ICreateTableCom
                     ;
 
                 table(createTable);
-                Execute(_commandInterpreter.CreateSql(createTable));
+                await ExecuteAsync(_commandInterpreter.CreateSql(createTable));
```
MikeAlhayek, line 54: @sebastienros I did not want to change existing logic in the PR. Maybe in another PR. but why do we need to create 3 different connections here? `Execute`, `CreateForeignKey`, and `AlterTable` all create different connection to the database. I think these 3 should be executed within a transaction on the same connection. Either way, even if we want to change this, I rather do that in a follow up PR just so we can track it separately in case we run into issues.

samples/YesSql.Samples.Web/Startup.cs
```diff
@@ -135,29 +135,28 @@ public async Task ConfigureAsync(IApplicationBuilder app, IWebHostEnvironment en
 
             var store = app.ApplicationServices.GetRequiredService<IStore>();
             store.RegisterIndexes(new[] { new BlogPostIndexProvider() });
-
-            using (var connection = store.Configuration.ConnectionFactory.CreateConnection())
+            Task.Run(async () =>
```
jtkech, line 138: Would need to be `awaited` (but in Configure() we can't use `await`) to not run in parrallel one task creating a table and another task using this table.

MikeAlhayek: @jtkech I added await on everything. This could maybe better placed in the controller or a new service like a migration

src/YesSql.Core/Services/DbBlockIdGenerator.cs
```diff
@@ -50,36 +51,40 @@ public async Task InitializeAsync(IStore store)
...
-
-                }
+                await transaction.CommitAsync();
+            }
+            catch
+            {
+            }
+            finally
+            {
+                await connection.CloseAsync();
```
sebastienros, line 74: Not necessary as `await using` will call `DisposeAsync`, which would then call `CloseAsync`.

src/YesSql.Core/Sql/SchemaBuilder.cs
sebastienros: Though, all these methods use the local `Connection` property, unless I missed something it's then using the same connection object.

src/YesSql.Core/Store.cs
```diff
@@ -91,115 +91,105 @@ public async Task InitializeAsync()
 
             if (!string.IsNullOrEmpty(Configuration.Schema))
             {
-                await using (var connection = Configuration.ConnectionFactory.CreateConnection())
-                {
```
sebastienros, line 95: Removing the code block means the connection will be disposed at the end of the method, and thus after the IdGenerator. That would lead to using more connections.

src/YesSql.Core/Services/DbBlockIdGenerator.cs
```diff
@@ -175,7 +184,8 @@ private void LeaseRange(Range range)
 
             } while (affectedRows == 0);
 
-            range.SetBlock(nextval, _blockSize);
+            await connection.CloseAsync();
```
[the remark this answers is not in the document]
MikeAlhayek: Yes. but isn't it better to release the connection before it is disposed?

samples/YesSql.Samples.Web/Startup.cs
MikeAlhayek: @sebastienros I added the awaiter

src/YesSql.Core/Services/DbBlockIdGenerator.cs
MikeAlhayek: removed

src/YesSql.Core/Store.cs
MikeAlhayek: added it back

samples/YesSql.Samples.Web/Startup.cs
sebastienros: Better option if you are willing to, change the project to use the new model that doesn't use a Startup file and is all async.

sebastienros:
Unit tests are failing since adding Rollabck in the DbGenerator, only on Sql, maybe the connection is already disposing the transaction. We need to understand what is happening. I am not sure we should just remove it since it's working for other databases. But that's a safety net.

sebastienros:
Are you getting lazy, or just bored?

MikeAlhayek:
> Are you getting lazy, or just bored?

I wish I was board and have nothing else better to do.

samples/YesSql.Samples.Web/Startup.cs
MikeAlhayek: @sebastienros I changed it to using Program.

MikeAlhayek:
@sebastienros I think all the issues are now fixed. 

- After modifying the `CommitOrRollbackTransactionAsync` logic, `await using var session = ...` works as expected and all tests now await and pass.
- In the `DbBlockIdGenerator`, I added the `RollbackAsync()` call you requested and it now worked. To get that to work, for some reason I had to set `throwOnError` to true when initializing `SchemaBuilder` instance.
- I also changed all the `BeginTransaction` calls to `BegingTransactonAsync`

Now, everything support async which is great! Let me know if you see anything else or if you good with merging this one.

sebastienros (approved):

src/YesSql.Core/Session.cs
```diff
@@ -630,7 +630,7 @@ public void Dispose(bool disposing)
             {
                 try
                 {
-                    CommitOrRollbackTransaction();
+                    CommitOrRollbackTransactionAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
jtkech, line 633: @sebastienros

Don't remember the details around this but are you sure about using `ConfigureAwait(false)` before `GetAwaiter()`?

MikeAlhayek: I think it just handles the error differently.

Do you see an issue?

jtkech: Would need more time to remember the details so not sure but reading again https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

For example this and that

> Using ConfigureAwait(false) to avoid deadlocks is a dangerous practice.

> The top-level methods, meanwhile, do require the context, so they cannot use ConfigureAwait(false)

I would not have used `.ConfigureAwait(false)` here but inside `CommitOrRollbackTransactionAsync()`, applied on the async call that it does.

Which may be useless if it is called asynchronously but maybe better for this synchronous usage.

But not sure ;)

jtkech: In the doubt I would not have used it at all.

MikeAlhayek: Okay. I'll remove it tomorrow before releasing the new version

MikeAlhayek: @jtkech what about in the other places like Session.Save and the SchemeBuilder?

jtkech: Idem for me

MikeAlhayek: Just to confirm, you suggest removing .ConfigureAwait(false) everywhere it's used? So .GetAwaiter().GetResult() only?

jtkech: Yes for now

MikeAlhayek: @jtkech done via https://github.com/sebastienros/yessql/pull/511 

Let me know if you need anything else modified

Issue #505, opened 2023-11-06
How do I make For<IIndex>().Map() support dynamic typing?


hyzx86:
for example, in the following code, I want for to support the base class and then return the object of the actual type in the Map

```c#
    public class CarIndex : MapIndex
    {
        public string Name { get; set; }
        public Categories Category { get; set; }
    }

    public class CarIndexProvider : IndexProvider<Car>
    {
        public override void Describe(DescribeContext<Car> context)
        {
            context
            .For<MapIndex>()
                .Map(c => new CarIndex
                {
                    Name = c.Name,
                    Category = c.Category,
                });
        }
    }
```

I am trying to use reflection implementation. It looks a bit complicated and difficult to understand. It would be much better if Yessql could support For(Type t)

![image](https://github.com/sebastienros/yessql/assets/15613121/60f3a3b2-829e-48a1-ab72-741695af227c)

hyzx86:
Maybe it has been supported at present, but I don't know how to use it yet.

hyzx86:
relate : https://github.com/sebastienros/yessql/pull/525

PR #521, opened 2023-12-25
Implement document change tracking

hyzx86:
Add a series of hooks to the session to monitor the update of the document object and implement actions after the document changes.

Such as implementing dynamic indexing, which seems to be difficult with the current design of Yessql.

In OC, neither ContentHandler nor a fake IndexProvider can determine whether an object needs to be saved to the database immediately. So the custom dynamic indexing functionality can only be written to the database immediately each time

```c#
            context.For<EocContentItemMapIndex>().When(contentItem =>
            {
                _serviceProvider = ShellScope.Services;
                UpdateIndex(contentItem).GetAwaiter().GetResult();
                return false;
            });
```

Through these hooks, we can insert custom code into the `_commands` list of the Session to achieve free insertion when the document is updated
 
relate:
https://github.com/sebastienros/yessql/issues/505


hyzx86:
@sebastienros Could you please help me to review this PR? thanks.

hyzx86:
Use another PR https://github.com/sebastienros/yessql/pull/525

PR #524, opened 2023-12-27
Add a way to specify no-tracking to reduce caching on demand

### LINKED ISSUE #522: Add a way to deattach multiple objects at the same time.

### LINKED ISSUE #523: Add a way to specify no-tracking in query time to prevent caching when not needed
Sometimes, we may want to query large data. We know that this data is for special purposes like reporting. So it does not need any internal caching. Internal caching can be very expensive when dealing with a large data set.

adding 
```
public class QueryContext
{
    public bool WithNoTracking { get; set; }
}
```

and passing this to the query can allow us to exclude the data from being cached.

MikeAlhayek:
Fix #523 
Fix #522


hishamco:
Can we add it as an extension method `AsNoTracking` or `WithNoTracking` instead of passing a context in each method

MikeAlhayek:
@hishamco I thought about it originally. The problem if we do that, we would have to track the state whether to track the entities per session not per query. The idea is to be able to not track the results per-query not per-session. Like what if the users wants to first query non-trackable documents, and in the same scope, they want to query other trackable documents? If they don't remember to set the state before the second query, both queries will not be tracked.

I simplified things by adding `WithNoTracking()` method of the `IQuery<T>` so if you are building a query, you can use this method to pass instruction to not track the entries.

But, in the Session, I think it is better to provide `QueryContext` that would change the behavior of the query on demand.

src/YesSql.Abstractions/IQuery.cs
```diff
@@ -111,6 +111,12 @@ public interface IQuery<T> where T : class
         /// Adds a Group By clause to the query for fold documents when an index returns multiple records.
         /// </summary>
         IQuery<T> NoDuplicates();
+
+        /// <summary>
+        /// When set, the fetched objects will not be cached internally.
```
hishamco, line 116:
```suggestion
        /// No tracking is involved for the fetched objects.
```

src/YesSql.Abstractions/IQuery.cs
```diff
@@ -111,6 +111,12 @@ public interface IQuery<T> where T : class
...
+        /// </summary>
+        /// <returns></returns>
```
hishamco, line 118:
```suggestion
/// <remarks> When called, the fetched objects will not be cached internally.</remarks>
```

hishamco (changes requested):
No looks much better for me, waiting for Seb comment

src/YesSql.Abstractions/IQuery.cs
MikeAlhayek: sorry I don't like the new comment.

MikeAlhayek:
> Now looks much better for me, waiting for Seb comment

If you spent time to review a PR, you may want to consider approving what you think is ready. This way, any changes you request will be to eventually to approve the PR which is more helpful than just leaving comment :)

hishamco:
Just fixing a small comment then I will approve

LGTM

src/YesSql.Abstractions/IQuery.cs
hishamco: No problem, at least remove un empty returns tag, also there's no set in the method

MikeAlhayek: @hishamco updated,

hishamco (approved):

sebastienros:
> Internal caching can be very expensive when dealing with a large data set.

I don't think it adds much more overhead since it's only adding the items to a collection, and "dealing with a large dataset" means you will also do that. You should just get rid of the session when you are done with it and you don't want to keep the references around.

However I can see why you would not want to have to create a new Session for every page of data you load, by convenience, and you came up with the notion of `QueryContext`. I think it should still be in the Session and not at the query level. This way you create a session that doesn't save the state and can reuse it to do all your queries. So I would suggest to just add a property to the session to disable tracking altogether. And another advantage is that the object won't even be added to the identity map at all, so potentially even faster processing. This would be an optional bool in `CreateSession`. I don't think it matter to be able to change the state of the session to enable/disable tracking dynamically. Just create a new session if you need different characteristics.

Also can you rename `DetachRange` to `Detach(IEnumerable<object>)` instead? Unless you have an argument that makes it better.

MikeAlhayek:
@sebastienros I made the requested changes.

PR #525, opened 2023-12-28
Allow describing an index from a given `Type`.

hyzx86:
Here is an example

```c#
public class PropertyDynamicIndexProvider : IndexProvider<Property>
{
    public override void Describe(DescribeContext<Property> context)
    {
        context
            .For(typeof(PropertyIndex))  // in addition to the existing ".For<PropertyIndex>()"
            .Map(property => new PropertyIndex
            {
                Name = property.Name,
                ForRent = property.ForRent,
                IsOccupied = property.IsOccupied,
                Location = property.Location
            });
    }
}
```


src/YesSql.Core/Commands/IndexCommand.cs
```diff
@@ -67,13 +67,19 @@ protected static void GetProperties(DbCommand command, object item, string suffi
 
         protected static PropertyInfo[] TypePropertiesCache(Type type)
         {
-            if (TypeProperties.TryGetValue(type.FullName, out var pis))
+            if (TypeProperties.TryGetValue(type, out var pis))
             {
                 return pis;
             }
+            //Avoid InvalidCastException
```
MikeAlhayek, line 74: New line before the comment. Single space after the //

test/YesSql.Tests/Indexes/PropertyIndex.cs
```diff
@@ -26,4 +26,20 @@ public override void Describe(DescribeContext<Property> context)
                 });
         }
     }
+    public class PropertyDynamicIndexProvider : IndexProvider<Property>
```
MikeAlhayek, line 29: New line before the class

test/YesSql.Tests/SqliteTests.cs
```diff
@@ -139,5 +139,50 @@ await builder
...
+
+                await builder
+                    .CreateMapIndexTableAsync<PropertyIndex>(column => column
+                        .Column<string>(nameof(PropertyIndex.Name), col => col.WithLength(4000))
+                        .Column<bool>(nameof(PropertyIndex.ForRent))
+                        .Column<bool>(nameof(PropertyIndex.IsOccupied))
+                        .Column<string>(nameof(PropertyIndex.Location), col => col.WithLength(4000))
+                    );
+
+                await builder
```
MikeAlhayek, line 164: I don't think this index will work on MySQL. Reduce the length from 4000 to like 200 on both properties to stay well under MySQL limits

MikeAlhayek (changes requested):

src/YesSql.Abstractions/Indexes/DescribeContext.cs
```diff
@@ -25,12 +25,19 @@ public IEnumerable<IndexDescriptor> Describe(params Type[] types)
...
+            return For<IIndex, object>(indexType);
+        }
         public IMapFor<T, TIndex> For<TIndex>() where TIndex : IIndex
         {
             return For<TIndex, object>();
         }
-
         public IMapFor<T, TIndex> For<TIndex, TKey>() where TIndex : IIndex
+        {
+            return For<TIndex, object>(null);
```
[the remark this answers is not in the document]
hyzx86: Although we can use default parameters, 
the For method without parameters will be thrown during the program run
```c#
  public IMapFor<T, TIndex> For<TIndex, TKey>(Type indexType = null) where TIndex : IIndex
  ```

File: src/YesSql.Abstractions/Indexes/DescribeContext.cs [hunk 5]
```diff
@@ -40,6 +47,12 @@ public IEnumerable<IndexDescriptor> Describe(params Type[] types)
             }
 
             var describeFor = new IndexDescriptor<T, TIndex, TKey>();
+
+            if (indexType != null)
```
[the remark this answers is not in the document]
hyzx86: > I don't think you need to check for null here. Maybe throw exception of the indexType argument is null.

if indexType is null , use default value : typeof(TIndex)

test/YesSql.Tests/SqliteTests.cs
hyzx86: But this unit test only for Sqlite  :)

src/YesSql.Abstractions/Indexes/DescribeFor.cs
```diff
@@ -51,8 +51,7 @@ public class IndexDescriptor<T, TIndex, TKey> : IDescribeFor, IMapFor<T, TIndex>
         private Func<object, bool> _filter;
 
         public PropertyInfo GroupProperty { get; set; }
-        public Type IndexType { get { return typeof(TIndex); } }
-
+        public Type IndexType { get; set; } = typeof(TIndex);
```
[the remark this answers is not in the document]
hyzx86: done

src/YesSql.Core/Commands/IndexCommand.cs
hyzx86: done

test/YesSql.Tests/Indexes/PropertyIndex.cs
hyzx86: done

README.md
```diff
@@ -7,6 +7,29 @@ A .NET document database interface for relational databases, because in SQL we (
...
+                {
+                    Name = property.Name,
+                    ForRent = property.ForRent,
+                    IsOccupied = property.IsOccupied,
+                    Location = property.Location
+                });
+        }
+    }
+```
+
```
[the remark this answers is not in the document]
hyzx86: done

src/YesSql.Abstractions/Indexes/DescribeContext.cs
```diff
@@ -25,12 +25,19 @@ public IEnumerable<IndexDescriptor> Describe(params Type[] types)
                 });
         }
 
+        public IMapFor<T, IIndex> For(Type indexType)
+        {
+            return For<IIndex, object>(indexType);
+        }
```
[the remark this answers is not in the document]
hyzx86: done

src/YesSql.Abstractions/Indexes/DescribeContext.cs
```diff
@@ -25,12 +25,19 @@ public IEnumerable<IndexDescriptor> Describe(params Type[] types)
...
         public IMapFor<T, TIndex> For<TIndex>() where TIndex : IIndex
         {
             return For<TIndex, object>();
         }
-
         public IMapFor<T, TIndex> For<TIndex, TKey>() where TIndex : IIndex
+        {
+            return For<TIndex, object>(null);
+        }
```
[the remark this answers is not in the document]
hyzx86: done

hyzx86:
Hi @MikeAlhayek  ， All problems resolved

test/YesSql.Tests/SqliteTests.cs
MikeAlhayek: @hyzx86 
No wonder it worked :) This test should be moved to CoreTests so it is executed on all providers.

Unit tests should pass on all database providers. It is easy to run Sqlite locally. But during deployment, we run tests on all providers.

During PR https://github.com/sebastienros/yessql/blob/main/.github/workflows/pr.yml#L61-L101

And during deployment https://github.com/sebastienros/yessql/blob/main/.github/workflows/publish.yml#L69-L100

MikeAlhayek: @hyzx86 Here is a proof that it failed MySQL ;) Fixing the issue in the next commit

image

MikeAlhayek (dismissed):
@hyzx86 I simplified things. I also removed the update you had to the readme file. That can be part of the Wiki once this PR is merged.

hyzx86:
@MikeAlhayek Thanks for your help . 
I didn't notice that this test should be moved to CoreTests because I copied it from one of the Sqlite Index tests above

MikeAlhayek:
I guess since it's just to ensure the index is created using type, we could also have left it in the SQLite only. It does not matter either way. It does not hurt running it on all databases either

src/YesSql.Core/Commands/IndexCommand.cs
```diff
@@ -67,20 +68,13 @@ protected static void GetProperties(DbCommand command, object item, string suffi
 
         protected static PropertyInfo[] TypePropertiesCache(Type type)
         {
-            if (TypeProperties.TryGetValue(type, out var pis))
+            if (TypeProperties.TryGetValue(type.FullName, out var pis))
```
hyzx86: This change is mainly to avoid a scenario
Because dynamically typed indexes are allowed in the code, there is no guarantee that the type will not change during use,
For example, when I save the type, it has 5 fields, and then I add a new field, then I need to regenerate the type index.
Two types of inconsistent errors occur


```log
2023-12-29 10:42:33.4084|SalesProtalDev|00-91aa5d5c1f96b614b2ebf3f2e6dfdbd5-5d2a9c17352de5d4-00||OrchardCore.ContentManagement.DefaultContentManager|ERROR|IContentHandler thrown from OrchardCore.Contents.AuditTrail.Handlers.AuditTrailContentHandler by InvalidCastException System.InvalidCastException: [A]EasyOC.DynamicIndex.AmisSchemaDIndex cannot be cast to [B]EasyOC.DynamicIndex.AmisSchemaDIndex. Type A originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array. Type B originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array.
at YesSql.Serialization.PropertyInfoAccessor.Invoker`2.Invoke(Object target)
at YesSql.Commands.IndexCommand.GetProperties(DbCommand command, Object item, String suffix, ISqlDialect dialect)
at YesSql.Commands.CreateIndexCommand.AddToBatch(ISqlDialect dialect, List`1 queries, DbCommand batchCommand, List`1 actions, Int32 index)
at YesSql.Session.BatchCommands()
at YesSql.Session.FlushAsync()
at YesSql.Session.FlushAsync()
at YesSql.Services.DefaultQuery.CountAsync()
at YesSql.Services.DefaultQuery.QueryIndex`1.YesSql.IQueryIndex<T>.CountAsync()
at OrchardCore.Contents.AuditTrail.Handlers.AuditTrailContentHandler.RecordAuditTrailEventAsync(String name, IContent content)
at OrchardCore.Modules.InvokeExtensions.InvokeAsync[TEvents,T1](IEnumerable`1 events, Func`3 dispatch, T1 arg1, ILogger logger) at YesSql.Serialization.PropertyInfoAccessor.Invoker`2.Invoke(Object target)
at YesSql.Commands.IndexCommand.GetProperties(DbCommand command, Object item, String suffix, ISqlDialect dialect)
at YesSql.Commands.CreateIndexCommand.AddToBatch(ISqlDialect dialect, List`1 queries, DbCommand batchCommand, List`1 actions, Int32 index)
at YesSql.Session.BatchCommands()

```

hyzx86: Because the index type may change while the program is running.

Just like this:

thrown from OrchardCore.Contents.AuditTrail.Handlers.AuditTrailContentHandler by InvalidCastException System.InvalidCastException: [A]EasyOC.DynamicIndex.AmisSchemaDIndex cannot be cast to [B]EasyOC.DynamicIndex.AmisSchemaDIndex. Type A originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array. Type B originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array.

hyzx86:
@MikeAlhayek I reupdated the code
``` diff
        protected static PropertyInfo[] TypePropertiesCache(Type type)
        {
            if (TypeProperties.TryGetValue(type, out var pis))
            {
                return pis;
            }

+            // If the same name type is already cached, it should be removed from the cache.
+            var existsingType = TypeProperties.Keys.FirstOrDefault(x => x.FullName == type.FullName);
+            if (existsingType != null)
+            {
+                var existsingProps = TypeProperties[existsingType];
+                foreach (var prop in existsingProps)
+                {
+                    PropertyAccessors.Remove(prop, out _);
+                }
+
+                TypeProperties.Remove(existsingType, out _);
+            }

            var properties = type.GetProperties().Where(IsWriteable).ToArray();
            TypeProperties[type] = properties;
            return properties;
        }
```

MikeAlhayek:
@hyzx86 why do you need to remove a previously store type?

hyzx86:
> @hyzx86 why do you need to remove a previously store type?

Because my dynamic index is generated according to the TypeDefinition in the OC, it may be updated many times during the run of the program, and if allowed for a long time, this static collection will cache many useless index types



```c#
        public override void Describe(DescribeContext<ContentItem> context)
        {
            var contentTypes = _contentDefinitionManager.ListTypeDefinitionsAsync().GetAwaiter().GetResult();
            foreach (var contentType in contentTypes.Where(x => !x.HasStereotype() && x.Parts.Any(x => x.Name == nameof(DIndexPart))))
            {
                var config = _dynamicIndexAppService.GetDynamicIndexConfigAsync(contentType.Name).GetAwaiter().GetResult();
                if (config != null)
                {
                    // get or create a dynamic index type 
                    var indexType = _dynamicIndexAppService.GetOrCreateDynamicIndexTypeAsync(config).GetAwaiter().GetResult();
                    context.For(indexType)
                        .When(contentitem => contentitem.ContentType == contentType.Name)
                        .Map((contentItem) =>
                    {
                        if (!contentItem.Published || !contentItem.Latest)
                        {
                            return null;
                        }
                        var table = _fsql.CodeFirst.GetTableByEntity(indexType);
                        var entity = contentItem.ToModel(config, indexType, table);
                        return (EocContentItemMapIndex)entity;
                    });
                }
            }
        }
    }
```

MikeAlhayek:
If the type already exists, it would just return it. Why would you need to remove it. Not sure I understand. Maybe you want to add a new test case to represent the problem.

hyzx86:
First of all, the type name doesn't change, and according to the rules in my code it might generate one of these type names

`public class {TenantName}.DynamicIndex.{ContentTypeName}`

However, whenever a field changes, a new c# type is generated based on this content type definition

At this point, the newly generated type is exactly the same as the type name cached in Yessql, but they are no longer the same type.

So the following exception is thrown when the property lookup is performed

```
InvalidCastException System.InvalidCastException: 
[A]EasyOC.DynamicIndex.AmisSchemaDIndex cannot be cast to 
[B]EasyOC.DynamicIndex.AmisSchemaDIndex. 
Type A originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' 
in the context 'Default' in a byte array. Type B originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, 
PublicKeyToken=null' in the context 'Default' in a byte array.
```

Full log :

```
2023-12-29 10:42:33.4084xxxx|ERROR|IContentHandler thrown from OrchardCore.Contents.AuditTrail.Handlers.AuditTrailContentHandler by InvalidCastException System.InvalidCastException: [A]EasyOC.DynamicIndex.AmisSchemaDIndex cannot be cast to [B]EasyOC.DynamicIndex.AmisSchemaDIndex. Type A originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array. Type B originates from 'DynamicTypesAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array.
at YesSql.Serialization.PropertyInfoAccessor.Invoker`2.Invoke(Object target)
at YesSql.Commands.IndexCommand.GetProperties(DbCommand command, Object item, String suffix, ISqlDialect dialect)
at YesSql.Commands.CreateIndexCommand.AddToBatch(ISqlDialect dialect, List`1 queries, DbCommand batchCommand, List`1 actions, Int32 index)
at YesSql.Session.BatchCommands()
at YesSql.Session.FlushAsync()
at YesSql.Session.FlushAsync()
at YesSql.Services.DefaultQuery.CountAsync()
at YesSql.Services.DefaultQuery.QueryIndex`1.YesSql.IQueryIndex<T>.CountAsync()
at OrchardCore.Contents.AuditTrail.Handlers.AuditTrailContentHandler.RecordAuditTrailEventAsync(String name, IContent content)
at OrchardCore.Modules.InvokeExtensions.InvokeAsync[TEvents,T1](IEnumerable`1 events, Func`3 dispatch, T1 arg1, ILogger logger) at YesSql.Serialization.PropertyInfoAccessor.Invoker`2.Invoke(Object target)
at YesSql.Commands.IndexCommand.GetProperties(DbCommand command, Object item, String suffix, ISqlDialect dialect)
at YesSql.Commands.CreateIndexCommand.AddToBatch(ISqlDialect dialect, List`1 queries, DbCommand batchCommand, List`1 actions, Int32 index)
at YesSql.Session.BatchCommands()

```

Type generated part of the code for reference

``` C#
  private Type CreateDynamicType(DynamicIndexConfigModel config)
  {
      // Define the namespace and assembly for the dynamic type
      string nameSpace = config.EntityInfo.NameSpace;
      string assemblyName = "DynamicTypesAssembly";
      // Create the dynamic assembly
      AssemblyBuilder assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(
          new AssemblyName(assemblyName), AssemblyBuilderAccess.Run);

      // Create a dynamic module in the assembly
      ModuleBuilder moduleBuilder = assemblyBuilder.DefineDynamicModule(nameSpace);
      // Define the type
      TypeBuilder typeBuilder = moduleBuilder.DefineType(config.EntityInfo.FullName, TypeAttributes.Public);

      typeBuilder.SetParent(typeof(EocContentItemMapIndex));

      // Add the attributes to the type
      var indexCols = config.Fields.Where(x => !x.Disabled && x.IsDefaultIndex)
                                   .Select(x => x.DbFieldName).ToList();
      var indexNames = new List<string>()
      {
          "ContentItemId", "DocumentId"//, "Published", "Latest"
      };
      if (indexCols.Any())
      {
          indexNames.AddRange(indexCols);
      }
      typeBuilder.SetCustomAttribute(new CustomAttributeBuilder(
                  typeof(EOCIndexAttribute).GetConstructor(new[] { typeof(string), typeof(string[]) }),
                  ["IDX_{tableName}_DocumentId", new[] { string.Join(",", indexNames) }])
              );
...
```

hyzx86:
Another option is to make the type cache of Yessql replaceable, so external code can define how to cache these types
This is better because external code knows exactly when it needs to clean them up; for example, the current scheme removes a ContentType definition and the dynamic index type associated with that type is cached in the collection forever

hyzx86:
Well, I reproduced this problem through unit tests, because in OrchardCore, the storage can be initialized multiple times

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Reflection.Emit;
+using YesSql.Indexes;
+
+namespace YesSql.Tests.Indexes
```
MikeAlhayek, line 7: Please put `;` at the end of the name space to save indentation.

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+            foreach (var item in dynamicTypeDef.Fields)
+            {
+                BuildProperties(typeBuilder, item);
+            }
+            // Create the type
+            Type dynamicType = typeBuilder.CreateType();
+            // Return the dynamic type
+            return dynamicType;
+        }
+
```
MikeAlhayek, line 36: anywhere you see multiple new lines, please reduce the. To one new line.

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+        private static PropertyBuilder BuildProperties(TypeBuilder typeBuilder, DynamicField item)
+        {
+            var propertyName = item.Name;
+            var propType = item.FieldType;
+
+            // Define the property
+            PropertyBuilder propertyBuilder = typeBuilder.DefineProperty(propertyName, PropertyAttributes.None, propType, null);
+            FieldBuilder fieldBuilder = typeBuilder.DefineField("_" + propertyName, propType, FieldAttributes.Private);
+
+            // 使用 IL API 生成类型属性访问器
```
MikeAlhayek, line 48: Remove this comment

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+            // Define the getter method
+            MethodBuilder getterBuilder = typeBuilder.DefineMethod("get_" + propertyName,
+                MethodAttributes.Public | MethodAttributes.SpecialName | MethodAttributes.HideBySig, propType, Type.EmptyTypes);
+
+            ILGenerator getterIL = getterBuilder.GetILGenerator();
+            getterIL.Emit(OpCodes.Ldarg_0);
+            getterIL.Emit(OpCodes.Ldfld, fieldBuilder);
+            getterIL.Emit(OpCodes.Ret);
+
+
```
MikeAlhayek, line 58: Multiple new lines

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    public class DynamicTypeDef
+    {
+        public string NameSpace { get; set; }
+        public string ClassName { get; set; }
+        public IEnumerable<DynamicField> Fields { get; set; }
+    }
+    public class DynamicField
```
MikeAlhayek, line 82: Add new line

test/YesSql.Tests/Indexes/PropertyIndex.cs
```diff
@@ -24,6 +27,48 @@ public override void Describe(DescribeContext<Property> context)
...
+                                 prop.SetValue(obj, property.Location);
+                                 break;
+                         }
+                     }
+
+                     return (MapIndex)obj;
+                 });
+            }
+
+
```
MikeAlhayek, line 69: Multiple lines

test/YesSql.Tests/CoreTests.cs
```diff
@@ -6296,6 +6296,92 @@ await using (var connection = _store.Configuration.ConnectionFactory.CreateConne
...
+
+            await using var session2 = store2.CreateSession();
+            var testPropEntity = testProperties.FirstOrDefault();
+            await session2.SaveAsync(testPropEntity);
+            await session2.SaveChangesAsync();
+            var testProperties2 = await session2.Query<Property, PropertyIndex>().ListAsync();
+            Assert.NotEmpty(testProperties2);
+
+        }
+
```
MikeAlhayek, line 6382: Multiple lines

MikeAlhayek:
@hyzx86 
This is where we initialize the store in OC. Maybe try investigating why it is called more than once. Or add a state if the store in initialized or not before initializing again.


https://github.com/OrchardCMS/OrchardCore/blob/250c5ff4285c82e4e3ed6be49f7429005d0b0e11/src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs#L125

MikeAlhayek:
@hyzx86 you may also want to open an issue in OC if you think that the store is being initialized more than once. That is something we would need to address in OC not in YesSql.

If you are sure that multi-initialization happens, please provide a reproduction steps and if possible a small repo that would help debugging the issue. Ping me then, and I'll carve out time to investigate it.

src/YesSql.Abstractions/Indexes/DescribeContext.cs
```diff
@@ -25,21 +25,37 @@ public IEnumerable<IndexDescriptor> Describe(params Type[] types)
                 });
         }
 
+        public IMapFor<T, IIndex> For(Type indexType)
+        {
+            return For<IIndex, object>(indexType);
+        }
```
hishamco, line 31:
```suggestion
        public IMapFor<T, IIndex> For(Type indexType) => For<IIndex, object>(indexType);
```

test/YesSql.Tests/CoreTests.cs
```diff
@@ -6296,6 +6296,92 @@ await using (var connection = _store.Configuration.ConnectionFactory.CreateConne
...
+    
+
```
hishamco, line 6384:
```suggestion
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
hishamco, line 7:
```suggestion
namespace YesSql.Tests.Indexes;
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    public class DynamicTypeGeneratorSample
+    {
+        public static Type GenType(DynamicTypeDef dynamicTypeDef)
+        {
+            var nameSpace = dynamicTypeDef.NameSpace;
+            var indexTypeFullName = dynamicTypeDef.NameSpace + "." + dynamicTypeDef.ClassName;
+            var assemblyName = "DynamicTypesAssembly";
+            // Create the dynamic assembly
+            AssemblyBuilder assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(
```
hishamco, line 17:
```suggestion
            var assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+                new AssemblyName(assemblyName), AssemblyBuilderAccess.Run);
+
+            // Create a dynamic module in the assembly
+            ModuleBuilder moduleBuilder = assemblyBuilder.DefineDynamicModule(nameSpace);
```
hishamco, line 21:
```suggestion
            var moduleBuilder = assemblyBuilder.DefineDynamicModule(nameSpace);
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+            // Define the type
+            TypeBuilder typeBuilder = moduleBuilder.DefineType(indexTypeFullName, TypeAttributes.Public);
```
hishamco, line 23:
```suggestion
            var typeBuilder = moduleBuilder.DefineType(indexTypeFullName, TypeAttributes.Public);
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+                BuildProperties(typeBuilder, item);
+            }
+            // Create the type
+            Type dynamicType = typeBuilder.CreateType();
+            // Return the dynamic type
+            return dynamicType;
+        }
+
+
+
```
hishamco, line 38:
```suggestion

```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
hishamco, line 48:
```suggestion
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+            var propertyName = item.Name;
+            var propType = item.FieldType;
+
+            // Define the property
+            PropertyBuilder propertyBuilder = typeBuilder.DefineProperty(propertyName, PropertyAttributes.None, propType, null);
+            FieldBuilder fieldBuilder = typeBuilder.DefineField("_" + propertyName, propType, FieldAttributes.Private);
+
+            // 使用 IL API 生成类型属性访问器
+            // Define the getter method
+            MethodBuilder getterBuilder = typeBuilder.DefineMethod("get_" + propertyName,
```
hishamco, line 50:
```suggestion
            var getterBuilder = typeBuilder.DefineMethod("get_" + propertyName,
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+                MethodAttributes.Public | MethodAttributes.SpecialName | MethodAttributes.HideBySig, propType, Type.EmptyTypes);
+
+            ILGenerator getterIL = getterBuilder.GetILGenerator();
```
hishamco, line 53:
```suggestion
            var getterIL = getterBuilder.GetILGenerator();
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
hishamco, line 58:
```suggestion

```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+            getterIL.Emit(OpCodes.Ldarg_0);
+            getterIL.Emit(OpCodes.Ldfld, fieldBuilder);
+            getterIL.Emit(OpCodes.Ret);
+
+
+            // Define the setter method
+            MethodBuilder setterBuilder = typeBuilder.DefineMethod("set_" + propertyName,
```
hishamco, line 60:
```suggestion
            var setterBuilder = typeBuilder.DefineMethod("set_" + propertyName,
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+                MethodAttributes.Public | MethodAttributes.SpecialName | MethodAttributes.HideBySig, null, new[] { propType });
+            // Define the setter method
+            ILGenerator setterIL = setterBuilder.GetILGenerator();
```
hishamco, line 63:
```suggestion
            var setterIL = setterBuilder.GetILGenerator();
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
```diff
@@ -0,0 +1,87 @@
...
+            setterIL.Emit(OpCodes.Ldarg_0);
+            setterIL.Emit(OpCodes.Ldarg_1);
+            setterIL.Emit(OpCodes.Stfld, fieldBuilder);
+            setterIL.Emit(OpCodes.Ret);
+
+            // Set the getter and setter methods for the property
+            propertyBuilder.SetGetMethod(getterBuilder);
+            propertyBuilder.SetSetMethod(setterBuilder);
+            return propertyBuilder;
```
hishamco, line 72:
```suggestion

            return propertyBuilder;
```

test/YesSql.Tests/Indexes/DynamicTypeGeneratorSample.cs
hishamco, line 82:
```suggestion

    public class DynamicField
```

test/YesSql.Tests/Indexes/PropertyIndex.cs
```diff
@@ -24,6 +27,48 @@ public override void Describe(DescribeContext<Property> context)
...
+
+
```
hishamco, line 71:
```suggestion
```

hishamco (changes requested):
Format

hyzx86:
@MikeAlhayek OrchardCore Is a multi-tenant system, it will certainly initialize `IStore` many times, which is not the problem of OC.

I think the problem now is how to abstract the type cache, and then we can modify or define its implementation externally.

hyzx86:
> Format

@hishamco  Thanks, I think you sent a link to a code format specification, but I can't find it. Can you send it again? I must print it out 😁

hishamco:
> Can you send it again?

Hope I can find it and send it back to you

src/YesSql.Abstractions/IStore.cs
```diff
@@ -44,5 +44,6 @@ public interface IStore : IDisposable
         /// Returns the <see cref="ITypeService" /> instance used to create this store.
         /// </summary>
         ITypeService TypeNames { get; }
+
```
hishamco, line 47:
```suggestion
```

src/YesSql.Abstractions/Indexes/DescribeContext.cs
```diff
@@ -25,21 +25,34 @@ public IEnumerable<IndexDescriptor> Describe(params Type[] types)
...
         public IMapFor<T, TIndex> For<TIndex>() where TIndex : IIndex
         {
-            return For<TIndex, object>();
+            return For<TIndex, object>(typeof(TIndex));
         }
 
         public IMapFor<T, TIndex> For<TIndex, TKey>() where TIndex : IIndex
         {
+            return For<TIndex, object>(typeof(TIndex));
+        }
```
hishamco, line 38:
```suggestion
        public IMapFor<T, TIndex> For<TIndex, TKey>() where TIndex : IIndex => For<TIndex, object>(typeof(TIndex));
```

src/YesSql.Abstractions/Indexes/IndexTypeCacheProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public class IndexTypeCacheProvider
+    {
+        private static readonly ConcurrentDictionary<PropertyInfo, PropertyInfoAccessor> PropertyAccessors = new();
+        private static readonly ConcurrentDictionary<Type, PropertyInfo[]> TypeProperties = new();
+
+        public virtual PropertyInfoAccessor GetPropertyAccessor(PropertyInfo property)
+        {
+            return PropertyAccessors.GetOrAdd(property, p => new PropertyInfoAccessor(p));
+        }
```
hishamco, line 18:
```suggestion
        public virtual PropertyInfoAccessor GetPropertyAccessor(PropertyInfo property) => PropertyAccessors.GetOrAdd(property, p => new PropertyInfoAccessor(p));
```

src/YesSql.Abstractions/Indexes/IndexTypeCacheProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+        public virtual PropertyInfo[] GetTypeProperties(Type type)
+        {
+            if (TypeProperties.TryGetValue(type, out var pis))
+            {
+                return pis;
+            }
+
+            var properties = type.GetProperties().Where(IsWriteable).ToArray();
+            TypeProperties[type] = properties;
+            return properties;
```
hishamco, line 29:
```suggestion

            return properties;
```

src/YesSql.Core/Commands/IndexCommand.cs
```diff
@@ -15,16 +15,14 @@ namespace YesSql.Commands
 {
     public abstract class IndexCommand : IIndexCommand
     {
+        private const string _separator = ", ";
         protected const string ParameterSuffix = "_$$$";
+        protected static PropertyInfo[] KeysProperties = [typeof(IIndex).GetProperty("Id")];
+        private static IndexTypeCacheProvider IndexTypeCacheProvider;
```
hishamco, line 21:
```suggestion

        private static IndexTypeCacheProvider IndexTypeCacheProvider;
        protected static PropertyInfo[] KeysProperties = [typeof(IIndex).GetProperty("Id")];
```

test/YesSql.Tests/CoreTests.cs
```diff
@@ -6296,6 +6296,88 @@ await using (var connection = _store.Configuration.ConnectionFactory.CreateConne
...
+            {
+                Name = new string('*', 40),
+                IsOccupied = true,
+                ForRent = true,
+                Location = new string('*', 500)
+            };
+
+            await session.SaveAsync(property);
+            await session.SaveChangesAsync();
+            var testProperties = await session.Query<Property, PropertyIndex>(x => x.Id == 1).ListAsync();
```
hishamco, line 6360:
```suggestion

            var testProperties = await session.Query<Property, PropertyIndex>(x => x.Id == 1).ListAsync();
```

hishamco (changes requested):
One more time :)

hyzx86:
Dears ， I added a configuration class, `IndexTypeCacheProvider`  to abstract the implementation of the index type cache
It seems to be working well so far

I found another way to repeat or duplicate registration, which can also trigger this problem  with session.RegisterIndexes

MikeAlhayek:
@hyzx86 
> OrchardCore Is a multi-tenant system, it will certainly initialize IStore many times, which is not the problem of OC.

OC registers a singleton `IStore` for each tenant. During registering the `IStore` we register all indexes as you can see here 
https://github.com/OrchardCMS/OrchardCore/blob/ef1eaab35f571fe9500dfe9099a6154fef0a4c9e/src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs#L108-L114

so it wouldn't be creating multiple `IStore` instances. Maybe the issue is in something else.

hyzx86:
> @hyzx86
> 
> > OrchardCore Is a multi-tenant system, it will certainly initialize IStore many times, which is not the problem of OC.
> 
> OC registers a singleton `IStore` for each tenant. During registering the `IStore` we register all indexes as you can see here https://github.com/OrchardCMS/OrchardCore/blob/ef1eaab35f571fe9500dfe9099a6154fef0a4c9e/src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs#L108-L114
> 
> so it wouldn't be creating multiple `IStore` instances. Maybe the issue is in something else.



> @hyzx86
> 
> > OrchardCore Is a multi-tenant system, it will certainly initialize IStore many times, which is not the problem of OC.
> 
> OC registers a singleton `IStore` for each tenant. During registering the `IStore` we register all indexes as you can see here https://github.com/OrchardCMS/OrchardCore/blob/ef1eaab35f571fe9500dfe9099a6154fef0a4c9e/src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs#L108-L114
> 
> so it wouldn't be creating multiple `IStore` instances. Maybe the issue is in something else.

Not really, pls see  session.RegisterIndexes

MikeAlhayek:
@hyzx86 have you tried implementing `IIndexProvider` interface for indexes that use generic type instead of ‘IIndexProvider<T>`?

hyzx86:
> @hyzx86 have you tried implementing `IIndexProvider` interface for indexes that use generic type instead of ‘IIndexProvider`?

Yes, but eventually I gave up
The `TypeProperties` cache actually holds The result type of the map function

`IndexDescriptor<T, TIndex, TKey>.IndexType`   

It's only used when you delete an index, because the map index returns null when you delete an index, so you can't get its type

in `Store.cs`

Clearly limit the use of  `DescribeContext<>` Maybe we can use a factory method here and implement it in the interface

```
        private static Func<IDescriptor> MakeDescriptorActivator(Type type)
        {
            var contextType = typeof(DescribeContext<>).MakeGenericType(type);
            return Expression.Lambda<Func<IDescriptor>>(Expression.New(contextType)).Compile();
        }
```

MikeAlhayek:
If you use `IIndexProvider` then you would provide the type of index using `ForType()` I never tried it. But, this should work. From there you can caching the dynamic type not the type provided in `IIndexProvider<T>`

MikeAlhayek:
If you need to create `IndexTypeCacheProvider`, the default implementation should solve the problem you are facing. Meaning your comment have to be addressed by default otherwise you're code can't be accepted since you are adding API that allows generic,

> This interface needs to be reimplemented when dynamic type and multi-tenant scenarios are used,


You may want to add interface to the abstraction project maybe like this

```
public interface IIndexTypeCacheProvider
{
    PropertyInfoAccessor GetPropertyInfoAccessor(PropertyInfo property);

    PropertyInfo[] GetProperties(Type type);

    void Update(Type type);
}
```

The `IndexTypeCacheProvider` can go into the Core project.

But, I am not sure that the complexity added by `IndexTypeCacheProvider` is needed. Maybe @sebastienros can provide more feedback on the approach you taken.

hyzx86:
@MikeAlhayek  updated the default index cache key

Now , you can modify the test `PopulateIndexUsingGenericType`  

Comment on the following line

```c#
            // update index type cache
            store1.Configuration.IndexTypeCacheProvider?.UpdateCachedType(changedType);
```
And  run the test ，Errors will be reproduced


There is nothing wrong with using the full name of the type as the Key for the cache.

We can call it after regenerating the type
`IndexTypeCacheProvider. UpdateCachedType(Type)` updates the cached type

However, in an OC or other project , it may be better to replace it with an in-memory cache or a distributed cache, since both of these are tenant isolated in an OC.

And we can define the full name of the class to start with the tenant name when using dynamic type generation

src/YesSql.Abstractions/IConfiguration.cs
```diff
@@ -90,6 +91,13 @@ public interface IConfiguration
         /// Gets or sets the identity column size. Default is <see cref="IdentityColumnSize.Int32"/>.
         /// </summary>
         IdentityColumnSize IdentityColumnSize { get; set; }
+
+        /// <summary>
+        /// Index type cache provider 
+        /// This interface needs to be reimplemented when dynamic type and multi-tenant scenarios are used,
+        /// using tenant separated cache media insteadexTypeCacheProvider 
+        /// </summary>
+        IndexTypeCacheProvider IndexTypeCacheProvider { get; set; }
```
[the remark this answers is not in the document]
hyzx86: Sorry for taking so long to reply. Because I didn't submit my comments

hyzx86:
After a period of test run, it is found that this method is not reliable

Next I'm going to try to implement  `IIndexCommand` through external code

PR #544, opened 2024-05-02
Add Format and apply rules

MikeAlhayek:
This PR apply some cleanup.

Also adds char support for StartsWith, EndsWith, and Contains


src/YesSql.Core/Sql/SqlBuilder.cs
```diff
@@ -425,23 +425,24 @@ public virtual string ToSqlString()
 
         public ISqlBuilder Clone()
         {
-            var clone = new SqlBuilder(_tablePrefix, _dialect);
-
-            clone._clause = _clause;
-            clone._table = _table;
-
-            clone._select = _select == null ? null : new List<string>(_select);
```
sebastienros, line 433: Looks suspicious, revert

test/YesSql.Tests/CoreTests.cs
```diff
@@ -1966,19 +1966,19 @@ await using (var session = _store.CreateSession())
             await using (var session = _store.CreateSession())
             {
                 Assert.Equal(2, await session.Query().For<Person>()
-                    .With<PersonByName>(x => x.SomeName.StartsWith("S"))
```
sebastienros, line 1969: This is not supported, only StartsWith(string) is bound

sebastienros: This could be added, but will require a custom test too, see `static DefaultQuery()`

sebastienros:
```
Error Message:
   System.InvalidOperationException : Operation is not valid due to the current state of the object.
  Stack Trace:
     at System.Reflection.RuntimeMethodInfo.GetGenericMethodDefinition()
```

Probably due to the changes methods (string to char) that I mentioned. It can't find the method (the one accepting a char) and tried to get the generic form but fails

In `DefaultQuery` there is `MethodMappings.TryGetValue(methodInfo.GetGenericMethodDefinition(), out action)` which should check if it's a generic method before.

You still need to use the "string" based one

PR #567, opened 2024-05-30
Export Document Version Property

hyzx86:


hyzx86:
@MikeAlhayek  Please approve the workflow check

hishamco:
What's the issue that you are trying to resolve?

hyzx86:
> What's the issue that you are trying to resolve? 

The Version field in the exported database is used to do a sync lock check, e.g. there is now a sync lock in Yessql but not in OC
I want to reuse this field and logic for implementing sync locks in OC to ensure that data submitted by a user is not overwritten by another user

sebastienros:
> The Version field in the exported database is used to do a sync lock check, e.g. there is now a sync lock in Yessql but not in OC
I want to reuse this field and logic for implementing sync locks in OC to ensure that data submitted by a user is not overwritten by another user

I am confused because it was added in YesSQL for OC and we do use it in OC already, example:
https://github.com/OrchardCMS/OrchardCore/blob/53ce9d79fa09b7e0bbce147e0a4e5ecf90984633/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Stores/OpenIdTokenStore.cs#L551-L563

hyzx86:
Yes, I understand, but I'm talking about a situation where a user is updating content from a page.
For example, if two people open a page at the same time, and user A opens it first, and B opens it later, and then B submits it, and B submits it after that, then B's submission should be overwritten.

sebastienros:
This is exactly the scenario that is covered in my example. By default the second user overwrites the changes from the first user. With the flag on SaveAsync then it will be detected and the second user gets an exception.

sebastienros:
It's called optimistic concurrency.

hyzx86:
> This is exactly the scenario that is covered in my example. By default the second user overwrites the changes from the first user. With the flag on SaveAsync then it will be detected and the second user gets an exception. 

However, although we currently have `ContentItemVersionId` it is not continuous.
if we submit from a web page, we encounter the following scenario

1. the first user opens the edit page, and the version number is 1
2. the second user opens the edit page and submits it, the version number is 2
3. the third user opens the page, gets version 2, modifies it and submits it, and the version number is 3.
4. the first user finally submits, checking for version expiration.

hyzx86:
This example is not very appropriate. Can't think of a suitable example for a while 🤣

PR #568, opened 2024-06-02
format ConcurrencyException message

hyzx86:
Add more message for ConcurrencyException


MikeAlhayek (approved):

src/YesSql.Abstractions/ConcurrencyException.cs
```diff
@@ -4,7 +4,8 @@ namespace YesSql
 {
     public class ConcurrencyException : Exception
     {
-        public ConcurrencyException() : base("The document could not be updated as it has been changed by another process.")
+        public ConcurrencyException(Document document) :
```
sebastienros, line 7: What about exposing the Document in a property of the exception? Maybe the caller could then extract anything they want from it, and could also log a precise error.

hyzx86: That will be greate !, 

Because I did find some issues that were thrown out so far, it may have been rolled back, so I can't see the corresponding data content.

src/YesSql.Abstractions/ConcurrencyException.cs
```diff
@@ -1,11 +1,27 @@
...
+            Document = document;
+
+            _message= $"""
+The document could not be updated as it has been changed by another process.
+     ID: {Document.Id}
+   Type: {Document.Type}
+Version: {Document.Version}
+Content: {Document.Content}        
+""";
         }
```
hyzx86, line 23: I formatted the error messages and wrote them directly to the logs, which shouldn't need to be processed by an external application

PR #596, opened 2024-11-14
Add DetachAll and ResetAsync

sebastienros:


src/YesSql.Core/Session.cs
```diff
@@ -240,6 +240,31 @@ public void Detach(IEnumerable<object> entries, string collection)
...
+            CheckDisposed();
+
+            var state = GetState(collection);
+
+            foreach (var item in state.IdentityMap.GetAll())
+            {
+                state.Saved.Remove(item);
+                state.Updated.Remove(item);
+                state.Tracked.Remove(item);
+                state.Deleted.Remove(item);
```
deanmarcussen, line 254:
```
      state._concurrent?.Clear();
      state._saved?.Clear();
      state._updated?.Clear();
      state._tracked?.Clear();
      state._deleted?.Clear();
```

might allocate less, as the state may not have created those hashsets, i.e. you might be creating them just by trying to remove an item.

but minor.

src/YesSql.Core/Session.cs
```diff
@@ -970,6 +995,10 @@ private async Task ReleaseTransactionAsync()
                 state._tracked?.Clear();
                 state._deleted?.Clear();
                 state._maps?.Clear();
+
+                // Clear the identity map as we don't want to return stale data after committing some changes.
```
deanmarcussen, line 999: maybe remove the comment on line 991 now ;)

deanmarcussen (approved):
ty

PR #601, opened 2024-11-27
Throw a when two objects represent the same document in a session

sebastienros:


sebastienros:
With the recent changes on identity map, when `SaveChangesAsync()` is invoked to commit the transaction we clear the IM such that the next queries get refreshed data from the db.

With that change, if we kept a previously loaded object (and potentially modified) but also did a query that was loading the same identity, we could get two objects with different states. This will throw a meaningful exception now.

This happens in OC where the identity stores call `SaveChangesAsync` after each operation, committing the transaction, where they should do nothing and let auto-flush "flush" the db (without committing the transaction and clearing the IM), or call `FlushAsync()`. So in a parallel PR I will remove these calls and replaces them with `FlushAsync`.

src/YesSql.Core/Session.cs
```diff
@@ -128,6 +128,15 @@ public async Task SaveAsync(object entity, bool checkConcurrency = false, string
 
                 if (id > 0)
                 {
+                    // If we got an object from a different identity map, without change tracking, the object reference could be different than
+                    // the one in the identity map, and the previous state.IdentityMap.TryGetDocumentId would have returned false.
+                    // In this case we need to assume it's an updated object and not try to "Add" it to the identity map.
+
+                    if (state.IdentityMap.TryGetEntityById(id, out var _))
+                    {
+                        throw new InvalidOperationException("An object with the same identity is already part of this transaction. Reload it before doing any changes on it.");
```
MikeAlhayek, line 137: How would this code fix the problem we noticed today? Wouldn't this throw a different spin exception but still fails?

deanmarcussen (approved):

PR #627, opened 2025-04-07
Introduce cancellation token

deanmarcussen:
Introduces a cancellation token property on the async methods.

It flows this cancellation token to the underlying library, rather than trying to manage it, itself.

So `CancelAsync` behaviour remains the same.

Will add some tests tomorrow, just wanted to see how whether the current tests pass.

I added the Meziantou.Analyzer package, as it has a useful anaylzer as to whether all the cancellation tokens have been passed to all the methods that support them. Had to fix a few other things it suggested, and ignore a bunch of other warnings.


samples/YesSql.Samples.Performance/Benchmarks.cs
```diff
@@ -254,15 +254,21 @@ public async Task CreateUsersAsync()
 
                 if (batch % batchSize == 0)
                 {
-                    users.ForEach(async u => await session.SaveAsync(u));
+                    foreach(var user in users)
+                    {
+                        await session.SaveAsync(user);
+                    }
```
hishamco, line 260:
```suggestion
                    }

```

samples/YesSql.Samples.Performance/Benchmarks.cs
```diff
@@ -254,15 +254,21 @@ public async Task CreateUsersAsync()
...
                     session = _store.CreateSession();
                     users = new List<User>();
                 }
             }
 
-            users.ForEach(async u => await session.SaveAsync(u));
+            foreach(var user in users)
+            {
+                await session.SaveAsync(user);
+            }
```
hishamco, line 271:
```suggestion
            }

```

hishamco (approved):

PR #643, opened 2025-07-21
Fix Session.Delete() method.

### LINKED ISSUE #642: Session.Delete() should delete in memory object
When Delete() method is called, we should attempt to also delete the object from the memory incase it was added by other methods like SaveAsync().

MikeAlhayek:
Fixes #642

### DIFF
--- a/src/YesSql.Core/Session.cs
+++ b/src/YesSql.Core/Session.cs
@@ -59,10 +59,7 @@ namespace YesSql
         {
             foreach (var indexProvider in indexProviders)
             {
-                if (indexProvider.CollectionName == null)
-                {
-                    indexProvider.CollectionName = collection ?? string.Empty;
-                }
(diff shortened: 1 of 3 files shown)


src/YesSql.Core/Session.cs
```diff
@@ -474,6 +471,10 @@ public void Delete(object obj, string collection = null)
 
             var state = GetState(collection);
 
+            // First detach the object from the session incase it was previously added by other
+            // methods like SaveAsync or UpdateEntityAsync.
+            DetachInternal(obj, state);
```
sebastienros, line 476: Maybe it should distinguish between a brand new item and and an updated one, because the with a brand new one it should not be added to the Delete collection. This would also ensure that adding it again after the Delete call would work (test to be added).

The test should also check that the indexes are not computed if the object is forgotten.

PR #647, opened 2025-07-31
Run tests on all databases even if failing

sebastienros:


sebastienros:
@gvkries only on sqlite ... maybe the test is incomplete

sebastienros:
And there is no exception, which the issues are mentioned, just a failed Assert

PR #653, opened 2025-08-29
Reduce allocation

MikeAlhayek:


gvkries:
I don't think the ordering should be removed here - it’s explicitly documented in the interface to return items in a defined order. Removing it would break that expectation and potentially lead to unintended behavior for consumers relying on that contract.

MikeAlhayek:
The contract does not promise ordering, thus should not sort anything. 

Sorting should be handled by the consumer "if desired". But, now always sorted.

PR #701, opened 2026-04-09
Add Test for Enum stores as sting

MikeAlhayek:
@sebastienros if you agree with this fix, we'll need to merge this and ship 5.5 at it could cause a binary breaking change.

### DIFF
--- a/src/YesSql.Abstractions/IIndexColumnTypeAccessor.cs
+++ b/src/YesSql.Abstractions/IIndexColumnTypeAccessor.cs
@@ -1,11 +1,12 @@
 using System;
 
-namespace YesSql
+namespace YesSql;
+
+public interface IIndexColumnTypeAccessor
 {
-    public interface IIndexColumnTypeAccessor
-    {
-        void SetIndexColumnType(Type indexType, string collection, string columnName, Type dbType);
-        bool TryGetIndexColumnType(Type indexType, string collection, string columnName, out Type dbType);
-        void RemoveIndexColumnTypes(Type indexType, string collection);
-    }
+    void SetIndexColumnType(Type indexType, string collection, string columnName, Type dbType);
+
+    bool TryGetIndexColumnType(Type indexType, string collection, string columnName, out Type dbType);
+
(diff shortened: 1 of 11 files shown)


gvkries:
This feels like using a sledgehammer to crack a nut. It effectively allows OC to keep using a string type for a specific index table column, which I consider a bug.

MikeAlhayek:
> This feels like using a sledgehammer to crack a nut. It effectively allows OC to keep using a string type for a specific index table column, which I consider a bug.

The ORM should be provider-agnostic. IMO, this is a bug that is related to a single provider (Postgres) which should be handled in the ORM. With this change, we can store enum in a string or int columns with no problem based on the user's preference.

gvkries:
> The ORM should be provider-agnostic. IMO, this is a bug that is related to a single provider (Postgres) which should be handled in the ORM. With this change, we can store enum in a string or int columns with no problem based on the user's preference.

If you create the index table with YesSql without forcing a different column type (e.g., `string`), everything works correctly across all providers. Using the enum type as intended results in an integer column being created:

```csharp
await builder.CreateMapIndexTableAsync<CarIndex>(column => column
    .Column<string>(nameof(CarIndex.Name))
    .Column<Categories>(nameof(CarIndex.Category))
);
```

Using `string` for the enum only happens to work with some providers due to implicit type conversions in their database engines. This behavior is provider-specific and incidental. In my opinion, this is not a bug in YesSql.

sebastienros:
This is a breaking change since existing databased have stringified integers in them and this PR will now check for the actual string version of the enum in queries.

In order to keep working with existing databases we would need a flag:

EnumBehavior
- BadOldIntInColumn Int32 stored in a VARCHAR
- EnumString String in VARCHAR
- EnumInt32 Int32 in an INT

This PR flows the database type so the correct conversion can be done. Using a flag means we don't need to retain this type along, we just follow the configuration.
Drawback is that it's a configuration for the store and not per property.

NBs: 
- the issue only happens for Index types that have a property with an enum type.
- the default values also are different ('' vs 0) 

If we wanted it at the field level, we could make it an attribute or a per-field configuration.

PR #718, opened 2026-06-25
Resolve document types by reflection instead of throwing KeyNotFoundException

sebastienros:
## Problem

`TypeService` keeps a reverse map (persisted type-name string -> CLR `Type`) that is populated **lazily and only forward**: an entry is added only when a type is first **written** or **queried by its exact type** within a process. The string indexer used when **reading** rows did a hard dictionary lookup and threw on a miss:

```csharp
public Type this[string s] => s == "dynamic" ? typeof(object) : nameTypes[s]; // KeyNotFoundException
```

`Session.Get<T>(documents)` resolves a row whose stored `Type` differs from the requested `T` via `Store.TypeNames[document.Type]`. Because `GetAsync<T>(ids)` selects rows purely by id (no type filter), a by-id read can load a row of a different document type whose CLR type has **not been touched yet this process**, and the lookup throws:

```
System.Collections.Generic.KeyNotFoundException:
  The given key 'OrchardCore.Templates.Models.TemplatesDocument, OrchardCore.Templates' was not present in the dictionary.
   at YesSql.Services.TypeService.get_Item(String s)
   at YesSql.Session.Get[T](...)
   at YesSql.Session.GetAsync[T](...)
```

The reverse map is per-store and empty after every restart, filling in only as each type is first saved/queried, so the failure looks intermittent ("only right after restart"), which matches the report.

Reported downstream at OrchardCMS/OrchardCore#19453.

## Fix

- `TypeService[string]`: on a miss, fall back to `Type.GetType(s)` and cache the result (including `null`) instead of throwing. All loaded assemblies resolve their full type names, so persisted rows deserialize correctly even before the type is first saved this process.
- `Session.Get<T>`: if a row's type still can't be resolved (`null`), skip the row (`continue`) instead of crashing the whole read.

## Tests

Added `SqliteTests.ShouldNotThrowWhenLoadingUnregisteredDocumentTypeById`: save a `Car`, then over the same database open a **fresh** store (empty type cache) and `GetAsync<Article>([carId])`. Previously this threw `KeyNotFoundException`; now the foreign row is skipped and the read returns empty.

### DIFF
--- a/src/YesSql.Abstractions/TypeResolutionException.cs
+++ b/src/YesSql.Abstractions/TypeResolutionException.cs
@@ -1,0 +1,25 @@
+using System;
+
+namespace YesSql
+{
+    /// <summary>
+    /// The exception that is thrown when the CLR type of a stored document cannot be resolved from
+    /// its persisted type name.
+    /// </summary>
+    public class TypeResolutionException : Exception
+    {
+        /// <summary>
+        /// Gets the persisted type name that could not be resolved.
+        /// </summary>
+        public string TypeName { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TypeResolutionException"/> class.
+        /// </summary>
+        /// <param name="typeName">The persisted type name that could not be resolved.</param>
+        public TypeResolutionException(string typeName) : base($"The type '{typeName}' could not be resolved. The assembly defining this type may not be loaded, or the type may have been renamed or removed. Register the type with 'IStore.TypeNames' so it can be resolved when its documents are loaded.")
+        {
+            TypeName = typeName;
+        }
+    }
+}
--- a/src/YesSql.Core/Services/TypeService.cs
+++ b/src/YesSql.Core/Services/TypeService.cs
@@ -49,9 +49,11 @@ namespace YesSql.Services
                 // The reverse map is populated lazily, only when a type is first written or
                 // queried by its exact type during the current process. When reading a row whose
                 // type hasn't been registered yet, fall back to resolving it through reflection
-                // instead of throwing a KeyNotFoundException. The resolved type (or null when it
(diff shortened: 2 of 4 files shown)


sebastienros:
I believe the issue can happen when the document's stored type name is a sub class or the type parameter used. Example: `Get<Animal>(1)` when document `1` has a type `Dog`. If `Dog` was not saved or loaded directly (`Get<Dog>`) before then the resolution will fail. But `TypeNames[typeof(Animal)]` is correctly invoked before the conversion, so the type was initialized, not the correct one.

Almost worse to return `null` because the element might be there but we just can't load it.

The possible solutions are
- specifically pre-register all the sub-types before usage such that the index is complete
- find all sub-types for `T` on `Get<T>` to pre-initialize it automatically

sebastienros:
Good points. Updated the PR accordingly:

- `Get<Animal>(id)` where the stored type is `Dog` now works: `TypeService` resolves the missing name through `Type.GetType` (caching the result), so the concrete `Dog` deserializes correctly even though only `Animal` was registered before the read.
- When the name genuinely can't be resolved (renamed/removed assembly, or a custom `SimplifiedTypeName`), it no longer returns `null` or throws `KeyNotFoundException` — it throws a new `TypeResolutionException` with a descriptive message, so the document is never silently dropped.
- Reverted the `null` skip in `Session.Get<T>`; the indexer now either returns a real type or throws.

For the "index is complete" angle, OrchardCore will pre-register the concrete sub-types it persists (companion change for #19453) so this exception isn't hit in practice; this PR just makes YesSql's failure mode correct and clear when a type really can't be resolved.

Added two SQLite tests: one asserting a resolvable foreign type loads/filters without throwing, one asserting `TypeResolutionException` for an unresolvable stored type name.

